PR #13, opened 2019-05-23
Update weapons.xml

Raphaellb:
addind correct level requirement for destruction weaponds


dudantas (commented):
Thanks for the contribution. 💪

PR #40, opened 2019-06-01
fix databse rewaid daily history

granoob:


gpedro (commented):
the column `streak` is unused or `daystreak` is a replacement for him?
the default values for smallint and tinyint should be `0` (int) and not `'0'` (char).

please, check if other columns are been used.

PR #102, opened 2019-06-02
Fix old bug

granoob:
um bug antigo na source monster


dudantas:
Already have this correction in down of your changes

PR #106, opened 2019-06-03
Loyalty system by Petardako

Olimpotibia:
Give x % of total skills to all account characters if he have loyalty points in database colum.

Credits to me.

### DIFF
--- a/data/creaturescripts/creaturescripts.xml
+++ b/data/creaturescripts/creaturescripts.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <creaturescripts>
+	<!-- loyalty Points -->
+	<event type="login" name="loyalty" script="loyalty.lua"/>
 	
 	<!-- The Masters Voice Quest-->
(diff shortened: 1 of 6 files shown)


gpedro:
in your code the bonus is hardcoded for 5% (loyalty 360 points)

> a druid with magic level 89 and 5% loyalty bonus might have magic level 90 displayed in the skills window. 
> e.g. a knight with 100 base sword fighting (103 after loyalty bonus) will still gain skills as if they had 100 sword fighting.
>  for example a paladin with 100 distance will not gain 25 (25% of 100) but will more likely gain 3-4 skill

reference: https://tibia.fandom.com/wiki/Loyalty_System

Loyalty Points | Bonus
-- | --
360 | 5%
720 | 10%
1080 | 15%
1440 | 20%
1800 | 25%
2160 | 30%
2520 | 35%
2880 | 40%
3240 | 45%
3600 | 50%

Olimpotibia:
Need fix in player.lua to display rank ingame..

PR #107, opened 2019-06-03
(WIP) fix: itens with name and plural attribute with lowercase value

gpedro:
WIP - Working in Progress... please don't merge.


gpedro:
this is not a good idea. some items use uppercase (like roman numbers). needs are reviewed manually.

PR #145, opened 2019-06-07
Fix showUseHotkeyMessage to check item type

granoob:
OTland tfs


gpedro:
cherry-picked from otland/forgottenserver#2604

PR #156, opened 2019-06-10
duplicate storage

granoob:
esta storages estao duplicadas


gpedro (changes requested):
Have you changed any ocurrence of this usage?

gpedro:
after review, I read the full code and this is not applicable PR. Is the same storage but is used for different purposes.

`Global.setStorageValue` data is saved on `global_storage` table and `Player.setStorageValue` is saved on `player_storage`.

PR #158, opened 2019-06-12
chore: switching name to Angry Demons

Raphaellb:
switching to Angry Demons


gpedro:
thanks for your contribution <3.

PR #160, opened 2019-06-13
fix count item usable in hotkey - small useble pot delay

granoob:


luucassoares:
Tested, working!

dudantas:
@granoob thanks for the contribution, merged

PR #161, opened 2019-06-14
fix: bug on the cults of tibia doors

Raphaellb:
npc angelo adds storage player:setStorageValue(Storage.CultsOfTibia.Life.Mission, 2) to player, and the door next to him was requiring the storage to be greater then 2 making it impossible for player to proceed. the correct way is for the quest door to check if storage is >= to (Storage.CultsOfTibia.Life.Mission, 2) so player can proceed anytime after.


dudantas:
Thanks for your contribution, I have modified some speaks of PT for EN

PR #168, opened 2019-06-18
Update doors.lua

Raphaellb:
Missing door script to allow players to pass quest door after Misguided storages are added by talking to Noozer and asking to pass.


dudantas:
merged, thanks for your contribution! =D

PR #172, opened 2019-06-19
 old function in which it causes fixed error

granoob:


dudantas:
I'll close, in case you do the review of the pull let me reopen

Issue #155, opened 2019-06-10
first dragon quest bug


Raphaellb:
**Describe the bug**
After starting the first dragon quest, I get a message of "your quest log has been updated" after killing any type of monster. 

**To Reproduce**
Go to Tamoril, start the first dragon quest, and kill any monsters after.
Note: include any configuration/script required to reproduce.

Raphaellb:
![image](https://user-images.githubusercontent.com/42285366/59197713-f62fea00-8b5f-11e9-8d93-857ab83fa536.png)

it looks like the  "your quest log has been updated" stopped after it went past 200 dragons, but the counter is still going after I kill any creature.

PR #173, opened 2019-06-20
fix: issue #155

FearLucien:
this will fix the current issue listed, but the rest of quest requires testing

### DIFF
--- a/data/creaturescripts/scripts/quests/first dragon/dragonsKill.lua
+++ b/data/creaturescripts/scripts/quests/first dragon/dragonsKill.lua
@@ -5,8 +5,8 @@ function onKill(player, target)
 	if player:getStorageValue(Storage.FirstDragon.Start) < 1 then
 		return true
 	end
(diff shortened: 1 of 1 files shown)


data/creaturescripts/scripts/quests/first dragon/dragonsKill.lua
```diff
@@ -1,18 +1,17 @@
...
+	if target:isPlayer() or target:getMaster() then
 		return true
 	end
 	if player:getStorageValue(Storage.FirstDragon.Start) < 1 then
 		return true
 	end
-	if not targetMonster:getName():lower() == 'dragon' then
-		return true
+	if player:getStorageValue(Storage.FirstDragon.DragonCounter) < 200 then
+		if target:getName():lower() == 'dragon' then
```
gpedro: `target: getName ()` should be first or on every kill event, he will check player storage.

gpedro (changes requested):

PR #186, opened 2019-06-21
addition: script for the gold converter

dudantas:
Gif :

Gold Converter


gpedro:
and about the `magic gold converter`?

dudantas:
> and about the `magic gold converter`?

this is it

PR #202, opened 2019-06-28
fix: barbarian quest, NPC Sinatuki

sonoxGit:
--		  testeQ = player:getStorageValue(Storage.BarbarianTest.Questline)
--		  print(testeQ)


dudantas:
thanks for the contribution

PR #209, opened 2019-06-30
fix: temporary solution for store kicks

westwol:
We can not consider this as a final solution.


gpedro:
lol. if is overrun, probably the server is trying to read more bytes then we have. my suggestion is just to disable gamestore to prevent crash.

westwol:
Indeed, I just discovered that packets on sending offers are not properly sent, causing an overrun disconection. I confirm this is not a real solution and might lead to other issues, so it is not recommended to use it on production.

Information so far:
Using tibia 11.8 or higher, if player opens Store, the client sends two consecutive packets that are supposedly being handled by server.
- 250 (open store)
- 251 (parse browse offers)

Using tibia 10.x, we get a single packet:
- 250 (open store)

gpedro:
@westwol can you send us this packets to try understand the new struct?

westwol:
The function producing the kicks is **sendShowStoreOffers**, going to keep on trying to refactor this packet structure in order to make it work as inteed.

Issue #218, opened 2019-07-03
Svargrond arena doors bug


coney4:
When you try open door it shows error

Lua Script Error: [Action Interface]
data/actions/scripts/quests/svargrond arena/arenaDoor.lua:onUse
data/actions/scripts/quests/svargrond arena/arenaDoor.lua:13: attempt to index global 'storages' (a nil value)
stack traceback:
        [C]: in function '__index'
        data/actions/scripts/quests/svargrond arena/arenaDoor.lua:13: in function <data/actions/scripts/quests/svargrond arena/arenaDoor.lua:1>

Raphaellb:
are you using the lastest distro+pack? I have already tested the svargrond arena, and everything worked correctly for me.

coney4:
yes
Problem is something with this
```
local storages = {
	[26100] = Storage.SvargrondArena.Greenhorn,
	[27100] = Storage.SvargrondArena.Scrapper,
	[28100] = Storage.SvargrondArena.Warlord
}
```
this was deleted in some commit

and this line needs to be fixed 
`local cStorage = storages[item.actionid]`

gpedro:
that's right. this commit 387cef00b2cb8cd556ca94f9909f6e51cae98527 removed this table

eduardodantas95:
fixed on https://github.com/opentibiabr/OTServBR-Global/pull/224

PR #224, opened 2019-07-04
Fix issue #218, Svargrond doors

Sleet37:
Adds the missing 'storages' local variable values that were removed on 387cef0


dudantas:
thanks for the contribution

PR #266, opened 2019-07-15
Inquisition quest

Raphaellb:
testing and fixing any issues with Inquisition quest. 

first problem I found is at mission 4. if player loses the special flask, there is no way to get another one, and making it impossible to complete quest. added the possibility to buy special flask at mission 4 in order to continue quest. if there is a need to put this on NPC trade, let me know and I can change merge.


Raphaellb:
I completed the quest, as far as I can tell, this is the only script bug I found. if this is acceptable, please merge.

ghost (approved):

PR #277, opened 2019-07-17
Removes "TFS-SDK" dependency from the build under Windows/Linux

Costallat:
To keep aligned with TFS, this change will remove the dependency of the TFS-SDK

The first build will take a lot of time as it was not possible to separate the packages from the x64 and Win32 version, so it will need to compile all packages for each version. It will take approximated 45 minutes for each version/cfg, but after that it will take just 3~4 minutes as it will keep the packages on the cache.

I don't have a Windows server to check if it is working, I am trusting that it is okay as we were able to compile and because I got the changes from TFS with a very few changes. Also I don't Visual Studio to check if it is working too.

Another thing is that I replaced the GMP package with the Crypto++, also aligned with changes made on TFS.


cys9211 (commented):
Dosn't work on Windows 7 64bit

Costallat:
Can you paste the error? I will try to download a Windows 7 VM to check, but that will take some time

Also, thanks for testing it!

Costallat:
Did you follow this wiki?

https://github.com/otland/forgottenserver/wiki/Compiling-on-Windows-%28vcpkg%29

I forgot to say that it need to install vcpkg and install the packages, if it works with that tutorial I will adapt it to here.

cys9211:
Sorry I'm idiot. Running on Old DLL files..
Thanks for help bro ❤️

gpedro:
LGTM!

gpedro:
staff checklist
- [x] setup travisci
- [x] setup coverity

gpedro:
@Costallat entra no discord. pls

ghost (approved):

Issue #283, opened 2019-07-17
Revert buggy AStarNodes::getBestNode


ghost:
**Describe the bug**
The current AStarNodes::getBestNode we have is based on a buggy code posted on otland that makes both players and monsters to diagonal walking to reach the destination and maybe even other bugs.

current code:
`
AStarNode* AStarNodes::getBestNode(const Position& targetPos)
{
	if (curNode == 0) {
		return nullptr;
	}

	int32_t diffNode = 0;
	int32_t best_node_f = std::numeric_limits<int32_t>::max();
	int32_t best_node = -1;
	for (size_t i = 0; i < curNode; i++) {
		const int_fast32_t Sx = nodes[0].x;
		const int_fast32_t Sy = nodes[0].y;
		const int_fast32_t Cx = nodes[i].x;
		const int_fast32_t Cy = nodes[i].y;
		int32_t SdiffX = std::abs(targetPos.x - Sx);
		int32_t SdiffY = std::abs(targetPos.y - Sy);
		int32_t NdiffX = std::abs(targetPos.x - Cx);
		int32_t NdiffY = std::abs(targetPos.y - Cy);
		diffNode = ((nodes[i].f+(((NdiffX-SdiffX)*5)+((NdiffY-SdiffY)*5)))+(std::max(NdiffX, NdiffY)*5));
		if (openNodes[i] && diffNode < best_node_f) {
			best_node_f = diffNode;
			best_node_f = nodes[i].f;
			best_node = i;
		}
	}

	if (best_node >= 0) {
		return nodes + best_node;
	}
	return nullptr;
}
`

default tfs code:
`
AStarNode* AStarNodes::getBestNode()
{
	if (curNode == 0) {
		return nullptr;
	}

	int32_t best_node_f = std::numeric_limits<int32_t>::max();
	int32_t best_node = -1;
	for (size_t i = 0; i < curNode; i++) {
		if (openNodes[i] && nodes[i].f < best_node_f) {
			best_node_f = nodes[i].f;
			best_node = i;
		}
	}

	if (best_node >= 0) {
		return nodes + best_node;
	}
	return nullptr;
}
`

**To Reproduce**
1. Login any decent speed character (even rotworms following you will do it)
2. Click on a stair or really any place that is not in a straight way to you
3. See the behavior

**Expected behavior**
The creature would move toward the objective normally

**What happens**
It goes diagonal walking even if there is no obstacle between it and the objective

[...]

PR #288, opened 2019-07-18
Revert pathfind

LukSrT:
Revert: https://otland.net/threads/optimizing-tfs-pathfinding.257868/
#283


Sleet37:
Why was it reverted?
Did it present any bugs?

Costallat:
Yes, and also to keep aligned with TFS it was removed for now, but if want to fix it, feel free to clone the repository and send a PR fixing it.

What i understood was that this was causing a lot of problems on the server and was not recommend to enable it in production servers.

PR #298, opened 2019-07-22
Create LICENSE.md

Costallat:
Add a License to the project

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.  We, the Free Software Foundation, use the
+GNU General Public License for most of our software; it applies also to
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -0,0 +1,674 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+Also add information on how to contact you by electronic and paper mail.
+
+  If the program does terminal interaction, make it output a short
+notice like this when it starts in an interactive mode:
+
+    <program>  Copyright (C) <year>  <name of author>
```
gpedro: missing some replacements, uhm?

Costallat: Ops, sorry, I used the Github and did "next next next finish".

Shame on me!

gpedro: hahahah

LICENSE.md
```diff
@@ -631,8 +631,8 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    <one line to give the program's name and a brief idea of what it does.>
-    Copyright (C) <year>  <name of author>
+    OTServBR-Global - A free and open-source MMORPG server emulator written in C++
+    Copyright (C) 2019  Opentibiabr
```
gpedro: `Opentibiabr` > `OpenTibiaBR`

about software license, we need to add a third-party license too?

Costallat: I'm not expert in this License world, but I did a search here and found that they say as long as we keep an equal or compatible license with the third-party software we don't need to include it, TFS uses GPLv2 and OTX doesn't have a license, but as it is TFS based, it should have also the GPLv2 license. 

I also checked the TFS license and saw that someone did the same mistake that I did here, they only added the license without changing the program name, author and year.

Costallat: OTX -> https://github.com/malucooo/otxserver-new

the other is under GPLv2

https://github.com/mattyx14/otxserver

PR #299, opened 2019-07-23
Fix 'buy' from NPC using the old chat system

Costallat:


data/npc/lib/npcsystem/modules.lua
```diff
@@ -1170,7 +1170,9 @@ if Modules == nil then
 		shop_npcuid[cid] = 0
 
 		local parentParameters = node:getParent():getParameters()
-		local parseInfo = {
+
+        local player = Player(cid)
+        local parseInfo = {
 			[TAG_PLAYERNAME] = Player(cid):getName(),
```
ghost: mudar aqui pra player:getName()?

PR #313, opened 2019-07-26
player.lua & login.lua update to calculate exp gain correctly & display 'XP Gain Rate' correctly

thatmichaelguy:
This pull request fixes the way bonus exp is applied and refactors the way in which we provide the XP Gain Rate value to the client.
As a part of this change, login.lua has has been updated to match and a switch has been added to startup.lua to allow hosts to choose if the rate/stage of the server should be included in the XP Gain Rate calculation.
I also disabled stages.xml by default and cleaned up the file.


data/events/scripts/player.lua
```diff
@@ -1012,95 +1012,50 @@ function Player:onGainExperience(source, exp, rawExp)
...
 		if (self:getPreyCurrentMonster(slot) == source:getName() and self:getPreyBonusType(slot) == CONST_BONUS_XP_BONUS) then
 			exp = exp + math.floor(exp * (self:getPreyBonusValue(slot) / 100))
 			preyTimeLeft(self, slot) -- slot consumption
 			break
 		end
 	end
-	
+
+	-- Store Bonus
+	useStaminaXp(self) –- Use store boost stamina
```
EPuncker: [Warning - Events::load] Can not load script: player.lua
data/events/scripts/player.lua:1037: '=' expected near '-'

miezaF:
Corrigir a linha 1037 do player.lua comentando.
Testado e aprovado.

ghost:
is it intended to display 7500%? because it would make more sense to get the base/stage rate to be 100% whatever the value is

thatmichaelguy:
I have done this intentionally yes, but we can change this fairly easily or provide a switch for the two options.
Personally I like the XP Gain Rate value to show the % increase based on 1x i.e. 10x = 1000%.
It's a nice way to show to the player the current exp rate, plus bonuses, so that they don't need to use e.g. !serverinfo

I asked the otland community what they thought:
https://otland.net/threads/client-xp-gain-rate-display.265467/

ghost:
yeah add a switch please and thank you

thatmichaelguy:
I've added the switch to config.lua and there really shouldn't be any issues but have not had a chance to actually test it yet.

majestyotbr:
LGTM

PR #317, opened 2019-07-27
Fix to exercise training

ghost:
- Removed some double / from file path in actions.xml and globalevents.xml
- Added sundial to watch.lua
- Added missing training weapons to all aplicable files
- Split training weapons into two groups, far use and close use
- Added exercise weapons to all respective NPCs
- Removed the vocation check, it is possible to use them wherever you like to, just like in cipbia (I've checked and confirmed it personally)
- Fixed its formula by @Flyckks
- Fixed typo on itemID 32402
- Custom training map now disabled by default until its fixed or removed


majestyotbr:
LGTM!

PR #346, opened 2019-08-01
Hero of Rathleton quest

Raphaellb:
Fix script issue preventing deep terror boss from spawning.


Raphaellb:
Completed quest 100%. only found the two script bugs above.

Issue #329, opened 2019-07-29
visual bug in magic level items


edilsonbim:
Items with ml do not add ml. just after relogin.
Tested on client 12

ghost:
they do add the ml, its only visual bug

edilsonbim:
already have a solution?

PR #374, opened 2019-08-08
Fix skill/maglevel display (client 12)

LukSrT:


ghost:
fixes #329

Sleet37:
How does this fix the issue?

LukSrT:
Magic level is now sent by skill, on client 12 and no longer by stats.

https://github.com/opentibiabr/OTServBR-Global/blob/develop/src/protocolgame.cpp#L3588

PR #376, opened 2019-08-09
tfs: apply patch that fixes the reload problem on Linux

Costallat:
Issue: #244 

- this fixes the crash bug with /reload creaturescripts on Linux


Znote:
Only on Linux? Is there any errors in Windows?

Costallat:
On Windows it works, but the patch should not affect that, it should work on both OS

Raphaellb:
tested using linux mint 18 with no problems.

miezaF:
relaod em linux 16.04 18.04 14.04 funcionando perfeitamente

PR #384, opened 2019-08-10
fix: falcon bastion tile and table

FakeShinoda:
- Changed some itens in falcon bastion
- Minor fixes arround the map
- Added few items in items.xml related to issue #383


majestyotbr:
LGTM

PR #392, opened 2019-08-11
minor changes to default values and settings

thatmichaelguy:
Personal opinion of how the base distribution should be.
- Lowered skill & magic rate in config.lua
- Disabled skill stages in skillstages.xml

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -118,9 +118,9 @@ classicAttackSpeed = false
 -- Rates
 -- NOTE: rateExp is not used if you have enabled stages in data/XML/stages.xml
 rateExp = 1
-rateSkill = 50
+rateSkill = 5
 rateLoot = 3
-rateMagic = 25
+rateMagic = 3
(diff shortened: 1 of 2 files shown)


oualid6496:
Why should this be merged this differs from person to person if this gets merged maybe tomorrow someone that prefers 100x will make a new pr for that. Won't that just go on and on?

thatmichaelguy:
I understand that rates are subjective, I do suggest we establish a default configuration that would play well "straight out of the box". 1x exp, 50x skills & 25x magic will not.

As for disabling skill stages - keeping it disabled by default seems logical to me. It'll just confuse new users if it is enabled by default.

Raphaellb:
I think we should leave it the way it is. this is trivial, and each person can modify to their liking.

majestyotbr:
I think this is a personal thing and each person can modify the options to their liking.
Anyway, thanks for the suggestion.

PR #394, opened 2019-08-11
fixed a typo and 2 npc's containg 2x buyable parameter

oualid6496:
fixed a typo
fixed 2x buyable parameter at the following npcs
- Edmund
- Valindara


majestyotbr:
Tested all NPCs in-game, no errors on distro. Thanks for the fix.

PR #396, opened 2019-08-12
Killing in the name of quest

Raphaellb:
quest fixes.


Raphaellb:
all done, these are the only script fixes it needed.

Raphaellb:
Pushed a couple more fixes for the Demon Oak quest, and the feyrist shrines were consuming wrong items
the ice and energy portal were reversed

the demon oak movement was not allowing players to enter without the holy icon
even after storage was added by entering it the first time with holy icon
if anyone wants to test.

majestyotbr:
LGTM

PR #405, opened 2019-08-13
Add mutex to control reload process

Costallat:
The server allows to run several reloads command but there are no control of when it finishes, so when we run the second reload, it will clean the same context that the first command is reading, and that will make the server crash.

To avoid that it was added an mutex to control the reload process and not let a second process to start before the first finish.

Tested here: Ubuntu 18.04

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -64,7 +64,7 @@ std::multimap<ScriptEnvironment*, Item*> ScriptEnvironment::tempItems;
 
 LuaEnvironment g_luaEnvironment;
 
-std::mutex m;
+std::mutex mutexReload;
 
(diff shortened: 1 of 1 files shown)


ghost:
isn't it better to add lets say, a 10 sec delay to every talk action instead?

Costallat:
The problem is that for me, it can happen right after I start the server if I run the reload creature scripts the server crash

gpedro:
if two gods run /reload, we will lock the lastest. timeout is not effective.

Costallat:
I tested the code again here and it is now working without the mutex to control the reload process, probably my branch was without the patch of the reload or I was with some local garbage, so I will close this PR.

PR #408, opened 2019-08-14
missing omrafir on monsters.xml

pitis91:
adding omrafir to the monsters.xml


miezaF:
Use lowercase on filename, linux doesnt reconize

PR #424, opened 2019-08-17
feat: add Battle, Ether and Zaoan Badger mounts

ghost:
Please update your client assets folder in order to use new content


majestyotbr:
Tested mounts, working. Clients 10 and 12 will be updated on [tools](https://github.com/opentibiabr/tools) soon.

PR #436, opened 2019-08-19
Fix Ocyakao raid to spawn Ocyakao, not Orshabaal

thatmichaelguy:
Funny surprise for a level 50 to find!


dudantas:
Thanks @thatmichaelguy 
Merged

PR #439, opened 2019-08-19
[TFS] Revscriptsys Monstertype event callbacks (#2594)

Costallat:
* Monster Type Event Callbacks

- now you can set the callbacks in the same file as you create the monster type in.

* removed unnecessary checks

I had to manually apply some changes, as our code has some changes that TFS doesn't have, so it is better to test this PR before merge.

Tks


EvilHero90:
there is a lot of missing stuff in this pr
https://github.com/otland/forgottenserver/pull/2594/files

Costallat:
> there is a lot of missing stuff in this pr
> https://github.com/otland/forgottenserver/pull/2594/files

Yep, I was checking that, I had to fix some files, and for some reason they disappeared after I push the changes.

gpedro:
@EvilHero90 looks great, uhm?

EvilHero90 (approved):

PR #446, opened 2019-08-20
New items and fixes

ghost:
- Fixed a lot of unknown item names, using both official and non-official info
- Added a few of 12.20 items
- Removed blank items


miezaF:
ue

PR #450, opened 2019-08-21
feat: analyzers by @Flyckks

gpedro:
original pull request made by  flyckks
credits to ggmartins9


techrlz18:
Why @Flyckks is receiving credits from this analyzers? This was created and published in github by Gagoosh from OTLand.

As you can see here:
https://otland.net/threads/tfs-1-2-tibia-11-analysers.264674/
https://github.com/ggmartins9/analysersv2

mattyx14:
The reference of the name mentioned here is why he was the one who exposed the thread with this code. And from there this has this tracking. Thanks for the observation.

PR #466, opened 2019-08-24
[WIP] Boosted Creature

miezaF:
With every server save, a randomly drawn creature will receive a temporary boost until next server save. During that time, the selected creature will:

- [X]  the list of creatures that can be picked is the same of prey lists

- [ ]  yield more experience points (double the normal amount).
- [ ]  carry more loot than usual (double the normal amount).
- [ ]  respawn at a faster rate (two times faster respawn rate than usual).
- [X]  The currently boosted creature will be displayed on the login screen of the client
- [X]  message on login:
- [X] List the raceId trough client 12+
- [X] list the monster name by raceId
- [ ] show monster name on login message
- [X] refresh boosted every startUp


data/creaturescripts/scripts/others/login.lua
```diff
@@ -85,14 +85,19 @@ end
 
 function onLogin(player)
 	local loginStr = 'Welcome to ' .. configManager.getString(configKeys.SERVER_NAME) .. '!'
+    local key = 56541
```
vankk, line 88: Please fix the tabs on thoses lines.

data/creaturescripts/scripts/others/login.lua
```diff
@@ -85,14 +85,19 @@ end
...
+    local resultId = db.storeQuery("SELECT `value` FROM `global_storage` WHERE `key` = " .. key)
+    local val = result.getString(resultId, "value")
```
vankk, line 90: Use `result.free` after get the results comming from db.storeQuery.

data/global.lua
```diff
@@ -9,6 +9,587 @@ STORAGEVALUE_ASSISTS = 30000
 STORAGEVALUE_DEATHS = 30001
 STORAGEVALUE_KILLS = 30002
 
+monsterIdName = {
```
vankk: Isn't better create a new file with thoses values and fix the indexs by incremental value?

data/globalevents/scripts/boosted.lua
```diff
@@ -0,0 +1,10 @@
+
```
vankk, line 1: Please remove this line.

data/globalevents/scripts/boosted.lua
```diff
@@ -0,0 +1,10 @@
...
...2", "509", "516", "555", "556", "557", "559", "560", "630", "693", "720", "723", "730", "733", "870", "872", "877", "898", "981", "1118", "1139", "1174", "1395", "1548", "1570", "1742", "120", "221", "511", "679", "881", "891", "919", "979", "982", "1042", "1141", "28", "81", "113", "114", "115", "119", "220", "258", "259", "290", "438", "439", "616", "620", "623", "624", "625", "627", "655", "656", "694", "695", "1096", "1525", "1817", "1818", "19", "265", "289", "513", "514", "515", "584", "1054", "1056", "1658", "18", "33", "37", "48", "65", "68", "78", "99", "100", "101", "256", "257", "281", "282", "283", "284", "286", "298", "321", "388", "446", "465", "483", "508", "512", "558", "580", "594", "595", "675", "696", "704", "707", "708", "710", "711", "712", "958", "959", "962", "975", "976", "1040", "1055", "1148", "1415", "1646", "1647", "1674", "1675", "1724", "1725", "1726", "1805", "26", "30", "36", "38", "43", "44", "45", "79", "82", "83", "124", "219", "251", "621", "631", "632", "633", "641", "674", "691", "709", "731", "732", "778", "786", "787", "788", "790", "791", "792", "796", "797", "801", "878", "899", "912", "1531", "1532", "1544", "1545", "1546", "1736", "1737",
```
vankk, line 4: Why don't use `monsterIdName` to get the values?

data/creaturescripts/scripts/others/login.lua
EvilHero90: It is a bad idea to query everytime a player login, this should just directly be saved in a table once it gets evaluated onStartup, there is actually no necessity to use sql overall

gpedro:
!!!
`login.php` cannot stay be here. this repository is focused on servergame, not website or client.

data/creaturescripts/scripts/others/login.lua
gpedro: @EvilHero90  that's right. onLogin is not the better place to do that. onStartup can be useful to update and some globalevent to refresh

PR #473, opened 2019-08-26
doCheckHouses commented out

thatmichaelguy:
I've commented out the doCheckHouses line in globalevents.xml
This is a useful script that we should retain, but do not want enabled by default.

I think it is okay to just leave it commented out.


Raphaellb:
yeah, I agree. thank you for finding that.

newacc12:
This is so users don't lose their houses if they don't log-in right? I do agree this is a good change. 

Will the users still lose their houses if they ran out of money from the bank?

PR #475, opened 2019-08-26
prey fix: data type compatibility.

westwol:


gpedro:
travis is reporting some error. can you fix? it's just `<< num <<` removal (i think)
```
/home/travis/build/opentibiabr/OTServBR-Global/src/iologindata.cpp:255:52: error: 
      use of overloaded operator '<<' is ambiguous (with operand types
      'basic_ostream<char, std::char_traits<char> >' and 'PreyState_t')
  ...query << player->getGUID() << ',' << num << ',' << PREY_STATE_SELECTION ...
```

Issue #469, opened 2019-08-25
command /save


genrider:
Why did you remove this command? This is very useful.

dudantas:
look on #476 
@genrider

PR #476, opened 2019-08-27
addition: talkaction /save

dudantas:
solves #469 
need tests


dudantas:
image

Tested

dudantas:
image
Tested

PR #480, opened 2019-08-27
[WIP] New loadMap Method

miezaF:


liontts:
Do we need this now, i mean, what are the benefits of changing or modifying something that isn't broken?

gpedro:
this is dismissible cuz we already have some similar system.

https://github.com/opentibiabr/OTServBR-Global/blob/305be216c55cbaefcdf6722054bb9421e250ba35/data/globalevents/scripts/custom/worldchanges/additionals/additionalsMapLoading.lua

PR #495, opened 2019-08-30
Added missing keywords to Lorek

newacc12:
Added missing keywords for mountain pass.
Required at least to be Ranger Rank to use Banuta, Chor and Mountain Pass.

https://tibia.fandom.com/wiki/Lorek


majestyotbr:
Thanks for the fix.

PR #497, opened 2019-08-30
Addons - Otimization

miezaF:
I use moses for 5 years so far.
Credits: https://github.com/Yonaba/Moses


gpedro:
sorry? that's for what?

gpedro:
for now, this will not be useful. we don't need to inflate engine to use a couple of functions.

PR #501, opened 2019-08-30
Fix Server Crash

FakeShinoda:
Fix crash when monster walk in any field.


Costallat (approved):

westwol (approved):

westwol (changes requested):
I would check if attackerPlayer is not nullptr
if (attackerPlayer && attackerPlayer->getPlayer()) {

majestyotbr:
Thanks for the fix.

Issue #461, opened 2019-08-22
Double respawn


newacc12:
**Describe the bug**
The feature to respawn monsters with players on the screen is not working correctly. Depending on the situation it can respawn new monsters even if the existing creature is still alive (double respawn).

**To Reproduce**
1. Go to the black knight quest (32874, 31947, 11) with a regular character.
2. Kill the beholders and scorpions only. Keep the black knight alive.
3. Stay on the SQM like in the screenshot below (32869, 31951, 11).
4. Wait until the new black knight respawn.

**Expected behavior**
Monsters don't respawn before they die.

**Screenshots**
![image](https://user-images.githubusercontent.com/20234557/63556288-cf9c1100-c509-11e9-9af9-2908ca16e9f6.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 16.04

**Additional context**
You can find additional information about this problem here: https://github.com/otland/forgottenserver/issues/2517 . I have posted this on the TFS forum last year as a question. On the original post I commented about a possible solution but I don't know if there would be any side effects to the respawn system. The TFS repository doesn't have this problem because it doesn't have this feature on its sources (or didn't at the time I posted).

majestyotbr:
https://github.com/opentibiabr/otservbr-global/commit/82fc51d1975124c1ac537151784af68a3c259890
https://github.com/opentibiabr/otservbr-global/commit/ebe63ef8c4d6a227af1a7f6859c5a5f0b3b96b5c

PR #502, opened 2019-08-30
Prevent Black Knight to Double Spawn based on SpawnRadius

miezaF:


newacc12:
@Flyckks I suppose this is to fix #461 right? Thanks for looking into it; however do you think this is the real fix? I pretty sure that this happens to other spawns as well (demons on the way to demon helmet is one example). I think we should somehow look into changing the source code to prevent this once for all. What do you think @MajestyOTBR ?

majestyotbr:
I agree with you @newacc12.

miezaF:
@newacc12 it should be applied to more places,  since The spawn should be based on Hunt radius, not monster pos radius.  So it become like 
monster pos = initial pos
radius = max dist to disappear or return to initial pos

newacc12:
@Flyckks I am not sure I get what you mean. As far as I understand, the radius on RME is meant to be used as a place where you can place monsters inside the map. I believe we should not increase the size just so we can fix an error in our source code that makes it spawn duplicated monsters. The despawnradius on the config file is what determine if the monster will disappear or not.

Just as an example, you 'fixed' the Black Knight by increasing the radius but on the same room, if you walk to the south, the server will start spawning duplicated bonelords. I don't see how this solution will be feasible to implement in all the respawns that we have. It seems that finding a solution on the source code will be much easier/better than changing the radius of all creatures.

I will play with the source code if I find the time and try to submit a PR so we can evaluate. Thanks for your input.

dudantas:
> @Flyckks I am not sure I get what you mean. As far as I understand, the radius on RME is meant to be used as a place where you can place monsters inside the map. I believe we should not increase the size just so we can fix an error in our source code that makes it spawn duplicated monsters. The despawnradius on the config file is what determine if the monster will disappear or not.
> 
> Just as an example, you 'fixed' the Black Knight by increasing the radius but on the same room, if you walk to the south, the server will start spawning duplicated bonelords. I don't see how this solution will be feasible to implement in all the respawns that we have. It seems that finding a solution on the source code will be much easier/better than changing the radius of all creatures.
> 
> I will play with the source code if I find the time and try to submit a PR so we can evaluate. Thanks for your input.

Thanks for the clarification.
In fact, you are absolutely right.
We should look for the root of the problem and not temporary solutions that will generate other bugs ...

PR #504, opened 2019-09-01
Updated Keys method

miezaF:
Use key, open door
It keep Opened
Use the key again, it locks.


liontts:
The power of a space is amazing!

gpedro:
is not about spacing. it's just some problems with rebasing and checkout

PR #505, opened 2019-09-01
Writing and product type correction

liontts:
Correction that allows the normal use of gold pouch and also correction in the item name.

The code itself works 100% according to my tests, it can be used immediately by simply revolving the comment.


majestyotbr:
Thanks for the fix.

PR #506, opened 2019-09-01
Fixes and updates

liontts:
* Setting values to default
* Correction in the name of new outfits
* Dropped customs tables


majestyotbr:
Thanks.

PR #514, opened 2019-09-02
fix: prey attempt 2

westwol:


liontts:
Hi, I just tested with the last two commits you sent about prey, well sometimes (not always) the player kicks when logging in, when I can log in when opening the system the player kicks the client instantly.

JoaoCRDias:
Tested this on lastest version os repo and didn't work

PR #516, opened 2019-09-02
Removing Unused Code

liontts:
The code is not required since the server database already has the accounts and characters.


gpedro:
ye, this is unuseful when you get website installation.

PR #517, opened 2019-09-02
Change to default gain values.

liontts:
The gain values of each vocation were changed, now they are default.


dudantas:
Thank you.
Our repository file is correct (in line with global tibia)

PR #520, opened 2019-09-03
[WIP]: Move raid system to globalevents

liontts:
"This effectively nukes the raid system in favor of global events, which permit better management for raids. It's easier to define raids on interval, on exact times, force execution, force scheduling, add monsters on different delays and randomized amounts and positions... and also gets rid of a command.

I added a lib with helper functions to add monsters to a raid. The event is called onRaid and there is an example bundled pretty equivalent to the previous XML raid example."

Credits to @ranisalt, code taken from the tfs repository.

I didn't have time to generate the raids, I hope someone takes the trouble to do it.


cellss:
Too many errors appears after I compiled with these changes...

PR #521, opened 2019-09-03
Fix exercise Wand/Rod

JoaoCRDias:
Considerar Magic Rate do server na exercise wand/rod

### DIFF
--- a/data/actions/scripts/others/objects/exerciseTraining.lua
+++ b/data/actions/scripts/others/objects/exerciseTraining.lua
@@ -15,6 +15,7 @@ local skills = {
 
 local dummies = {32142, 32143, 32144, 32145, 32146, 32147, 32148, 32149}
 local skillRate = 1*configManager.getNumber(configKeys.RATE_SKILL)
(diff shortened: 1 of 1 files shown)


thatmichaelguy:
We should update this as per https://tibia.fandom.com/wiki/User:Molx/Exercise_Efficiency
``One Exercise Rod or Exercise Wand is equivalent to drinking 600 Ultimate Mana Potions. This is also valid for Knights and Paladins.``
``The advancement of the main skills are the same for all vocations, i.e. it takes the same number of Exercise Weapons for mages to get Magic Level 50 than a Knight would need to get 50 Axe, Club or Sword Fighting or a Paladin would need to reach 50 Distance Fighting.``

1 UMP = ~500 MP
600 UMP = ~300,000 MP
300,000 / 500 tries = 600 MP/try ( * magicRate )

dudantas:
@Kolisium The idea is interesting, but I don't think it's the right way to handle it.
needs to be added to all skills and developed in a way that does not cause bugs or breaches.
If you want to continue the PR to bring the change to all skills, let me know that I reopen the PR.
Thanks.

PR #527, opened 2019-09-04
fix: bug in the talkactions

dudantas:
talkactions didn't work
now it's normal (I tested), but if anyone else wants to test


gpedro:
talkactions works great without this changes! please check if your `players.group_id` is defined to 6 and `accounts.type` to 5

oualid6496:
Its the account type since its working for him after he removed the account type checks. Like gpedro said change the accounts accounttype to 5.

dudantas:
@gpedro ok, thanks

PR #536, opened 2019-09-06
Migrating Events

liontts:
Migrating events to standalone scripting for better viewing also removed some unused events.


dudantas:
@liontts Thanks for the contribution.

PR #539, opened 2019-09-06
fix a problem with exercise weapons

lBaah:
Fixes a problem when a player logout while training with exercise weapons, which prevent the player to start training again. (#524 )


Sargitarium:
Bom dia nobre, acabei de testar as paradas das exercises e quando acabou as cargas, o char não pode mais usar.... tem mais alguma coisa ai a ser corrigida,
Se eu deslogar, elas voltam ao normal...

lBaah:
This should now work as intended. Thank you for the report.

majestyotbr (approved):

majestyotbr:

PR #540, opened 2019-09-06
Fixed some issues with assassin outfit quest

### LINKED ISSUE #529: Assassin Outfit Quest - Bug NPC Atrad
Hi!
Im testing all the outfit quests, and im having problems to talk with npc Atrad, the third npc of Assassin Outfit quest.When i try to talk, i get the error bellow:

bug

newacc12:
Added some keyworks to Atrad, Erayo and Vescu according to TibiaWikia.
Fixed console error when talking to Atrad.
Added the requirement to be invisible when visiting Erayo's hut.
Should fix #529

Can someone please let me know why [getCondition](https://github.com/opentibiabr/OTServBR-Global/compare/develop...newacc12:bug/assassin-outfit-quest?expand=1#diff-5235a6e7892151ed2fdba171f78be8b4R9) is returning nil when you have stealth ring equipped?

Reference: https://tibia.fandom.com/wiki/Assassin_Outfits_Quest


data/movements/scripts/quests/assassin outfit/vescuhouse.lua
```diff
@@ -0,0 +1,17 @@
+function onStepIn(creature, item, position, fromPosition)
+	local player = creature:getPlayer()
+
+	if not player then
+		return true
+	end
+
+	if position.x == 32519 and position.y == 32911 and position.z == 7 then
```
gpedro: cant you move position to variable outside function and just compare position. something like that:
```lua
local enterPosition = Position(32519, 32911, 7)

function onStepIn(creature, item, position, fromPosition)
-- ...

if position == enterPosition then
-- ... 
end
```

data/movements/scripts/quests/assassin outfit/vescuhouse.lua
```diff
@@ -0,0 +1,17 @@
...
+		local invisible = player:getCondition(CONDITION_INVISIBLE)
+		local slot = player:getSlotItem(CONST_SLOT_RING)
+		if not invisible and (not slot or not slot.uid == 2202) then -- This is ugly. I believe we should not have to check for the ring but for some reason getCondition is returning nil. Ideas?
```
gpedro: why the ring check is necessary? only invisible condition isn't enough?

gpedro (dismissed):

joseluis2g:
get condition needs a second parameter if you using it as a item, the slot parameter, add 9 as second param and it will check the ring slot

data/movements/scripts/quests/assassin outfit/vescuhouse.lua
newacc12: Totally! I will update it.

data/movements/scripts/quests/assassin outfit/vescuhouse.lua
newacc12: It was not working. I will try @joseluis2g suggestion.

newacc12:
> get condition needs a second parameter if you using it as a item, the slot parameter, add 9 as second param and it will check the ring slot

Thanks for the suggestion I will try it.

data/movements/scripts/quests/assassin outfit/vescuhouse.lua
dudantas: you can try this 
```suggestion
		if not invisible and (not slot or not slot.uid == 2202) then -- This is ugly. I believe we should not have to check for the ring but for some reason getCondition is returning nil. Ideas?
```

dudantas: you can try this 

```suggestion
function onStepIn(creature, item, position, fromPosition)
    if creature:isMonster() then 
        return false 
    end
    
    local player = creature:getPlayer()
    local enterPosition = Position(32519,32911,7)
    if position == enterPosition then
    if player:getItemById(2202, deepSearch) or player:getCondition(CONDITION_INVISIBLE) then
        else
        player:teleportTo(Position(32519, 32912, 7))
    end
    end
    
    return true
end
```

PR #545, opened 2019-09-06
chore: removal of some customs scripts

dudantas:


newacc12:
@dudantas While I do understand that the goal of this project is to keep as close as possible to global tibia, I think we should consider keeping the trainers. Nowadays, this is widely accepted and very popular functionality.

Thoughts?

dudantas:
I think it is very simple for those who have custom servers to create. I don't think it's necessary to be in the repository, because it's very easy to do that and I noticed that most people have their own custom trainers.

There are currently offline trainers and training dummy, I really don't see the need for custom trainers.
One thing is also that in the forums there is all this easily accessible content and easy installation.

PR #549, opened 2019-09-07
Workaroud for RSA server crash

lBaah:
Workaround to prevent the server to crash by logging in with a RSA key that is not expected by the server.

### DIFF
--- a/src/rsa.cpp
+++ b/src/rsa.cpp
@@ -31,9 +31,12 @@ static CryptoPP::AutoSeededRandomPool prng;
 
 void RSA::decrypt(char* msg) const
 {
-	CryptoPP::Integer m{reinterpret_cast<uint8_t*>(msg), 128};
-	auto c = pk.CalculateInverse(prng, m);
-	c.Encode(reinterpret_cast<uint8_t*>(msg), 128);
+	try {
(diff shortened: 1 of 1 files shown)


mrhultin:
Would it make any sense to log incorrect rsa key submissions? Havent had this happen so i am unsure how people use this to attack servers.

Looks solid, though I am not a fan normally of silent errors like this. However, I assume it never will be a issue for regular players.

gpedro:
i think this is wrong location to fix it. Should be better fix in [Protocol::RSA_decrypt](https://github.com/opentibiabr/OTServBR-Global/blob/609414b3686ae7fa033981ab6a9717cb223b63e8/src/protocol.cpp#L141) and force to close the connection like [Connection::parseHeader](https://github.com/opentibiabr/OTServBR-Global/blob/4fbafb7af2757204e18e6971cfc6acfdb9b3fa34/src/connection.cpp#L179-L181)

lBaah:
@mrhultin that doesn't make sense until you make a mistake and forget to change RSA key from the client. I did it and got the server crashed, that's how I found this bug. So if anyone make that mistake (including unadvised players), the server IS going to crash. 
Now that makes sense?

@gpedro I'll look into that. Thank you.

PR #550, opened 2019-09-08
added missing items to transform

### LINKED ISSUE #491: bath tub is not working as expected
**Describe the bug**
Transform.lua scripts isn't working as intended.

**To Reproduce**
For example let's take bath tub:
1. When 29312 is used (empty tub) it turns into 29313 (full tub).
2. When 29313 is used (full tub) it doesn't do shit.

**Expected behavior**
For example, the tubs 29312 and 29313 should be changed to the respective id when used.

**Screenshots**
Don't really think a screenshot of an empty/full tub will help here :P

Windows 10

**Additional context**
Some items do work and some don't the same goes for Demon Pet when faced to the right. Pretty sure there are more that don't work as intended haven't really tested them yet.

lBaah:
Added missing items from store to transform.lua

Should fix #491


majestyotbr:
LGTM

PR #551, opened 2019-09-08
Fixed a bug introduced by #540

newacc12:
Need brackets here otherwise this code will run for every sqm id 3139


majestyotbr:
Thanks for the fix.

PR #552, opened 2019-09-08
Applying ifdef to switch to use luajit/lua for windows and keeping others as lua.hpp

rafael-adcp:


dudantas:
Luajit is not necessary when you get the lua-cpp

But, thanks.

dudantas:
Luajit is not necessary when you get the lua-cpp

But, thanks.

PR #554, opened 2019-09-08
Added missing mage wand addon quest

newacc12:
This commit will also add the marriage system back (the code was already there). I am not sure why it was removed.


majestyotbr:
Thanks for your contribution.

PR #397, opened 2019-08-12
Golden outfit pull

pitis91:

PR #559, opened 2019-09-09
Feat: Golden Outfit

liontts:
Fix for golden outfit because the server has errors.
The code is not my own, but as it was in open-source project I felt free to postalo.

Credits: worthdavi

### DIFF
--- a/data/lib/miscellaneous/051-storages.lua
+++ b/data/lib/miscellaneous/051-storages.lua
@@ -638,6 +638,11 @@ Storage = {
 			AddonHatCloak = 12064
 		},
 
+		-- Golden Outfit
+		goldenOutfit = 11160,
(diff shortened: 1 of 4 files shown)


mrhultin:
Any reason you are not naming functions and variables in english? The text strings arent very well written. " This boots" should for example be "these boots"

mrhultin:
And redfine the same functions per npc instead of placing them in a lib file or such makes little to no sense. Repating code should never really be needed.

dudantas:
Good Morning!
The code is badly done, repeated and needs to be refactored. This code was originally posted here to the repository at https://github.com/opentibiabr/OTServBR-Global/pull/397
Also with error ...
How do you say? If it is to climb something bugged or poorly done, it is better not to.

Thanks.

PR #569, opened 2019-09-11
Fix #564

### LINKED ISSUE #564: Imbuiment System
Imbuiment system: mages just with the rod/wands can make Crititicals hits, can not produces only wiht spells, and all the time always is wasting, if you not using the set, the time don't stop to wasting.

joseamaya1:
Fix issue #564


majestyotbr:
Thanks for the fix.

PR #577, opened 2019-09-12
 [WIP] (SID to CID) Folder 'movements'

Costallat:
WIP!

But if you see that I skipped something, let me know so I can fix, it's hard to find all item id's inside some quest files.


dudantas:
Thanks ... But let's stop this development.

PR #579, opened 2019-09-13
updates exercise weapon calculations

lBaah:
Update on magic rate calculation and fix a typo.


cellss (commented):
Tested exercise wand from ML 0 with 25x ML rate:

From magic level 0 to 62 (39% to ml 63) using 500 charges.

Tested exercise wand from ML 0 with 50x ML rate:

From magic level 0 to 62 (33% to ml 63) using 500 charges.


Tested exercise club from skill 10 with Skill rate 50x:

From Skill Club 10 to 51 (75% to 52) using 500 charges.

Tested exercise sword from skill 10 with Skill rate 100x:

From Skill Club 10 to 51 (75% to 52) using 500 charges.

By the way, I think it's better than the current system

thatmichaelguy:
This is correct for exercise/training wands & rods.
Melee & Distance weapons still need to be updated though, but that shouldn't prevent this being merged.

PR #588, opened 2019-09-18
few first dragon quest fixes

Raphaellb:
Few first dragon quest fixes I could correct myself.


Raphaellb:
Update entranceTeleports.lua

This needs to be an "or" because only one of the two arguments needs to be true, and not both to prevent player from going into boss fight.

Update Gelidrazah'S Thirst.lua

replacing duplicate "yes" with questions in order to continue thru the quest to get storage.

Update 051-storages.lua

updating conflicting storages with data/actions/quests/first dragon/chests.lua

dudantas:
Thanks for contribution!

PR #589, opened 2019-09-18
Fix Docker script to start a test server due to changes on config.lua

Costallat:
- Fixes the docker scripts to start a local test server on linux environment due to changes on the config.lua file

- Re-enables otclient support on the server side


dudantas:
Thaks for contribution.

PR #590, opened 2019-09-18
[TFS] Store player direction(look direction)

Costallat:
Saves players direction on logout and restore on login.

(I mixed the code from the docker fix because I need it to test this, but once it get merged, I will fix this PR)


dudantas:
Please send a new PR with the direction look.

PR #591, opened 2019-09-18
TFS: Fix skull ticks overflow

Costallat:
●The reason for the change:
 
 When a player's skull time exceeds 35 days, the skull time overflows as it's stored as milliseconds which has no better value than seconds
 
 ●Reproducing Bug
 
 1-Let Player A gets 35 kills(which is equal of 35 days)
 2-35 days calculated in milliseconds -> 3024000000
 3-Maximum value for int32 -> 2147483648
 4-by calculating the difference = -876516352
 5-By next login the player will have 0 skull time
 6-Major problem Lied in [here](https://github.com/otland/forgottenserver/blob/master/src/player.cpp#L3828)
 
 ●Changes:
 
 Changed skull ticks from millisecond to seconds to avoid overflow and
 preserve performance because there's no need for using milliseconds.
 
 ●Test:
 1-Player A Killed Player B 35 times
 2-Player A now has a black skull
 3-Player A logged out showing the following information(skull time:3020086 seconds)
 4-Player A logged in after few seconds showing the following information(skull time:3020082 seconds)

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -86,7 +86,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `lastip` int(10) UNSIGNED NOT NULL DEFAULT '0',
   `save` tinyint(1) NOT NULL DEFAULT '1',
   `skull` tinyint(1) NOT NULL DEFAULT '0',
-  `skulltime` int(11) NOT NULL DEFAULT '0',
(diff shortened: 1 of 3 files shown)


westwol:
LGTM

mattyx14 (approved):

PR #592, opened 2019-09-18
fix hasCondition function

Costallat:
- this fixes the problem that hasCondition returned false if you used -1 in ticks for a condition (to make it infinite)


dudantas:
Thanks for the contribution.

PR #594, opened 2019-09-18
[revscriptsys] call clear() on action/movement vectors 

Costallat:
- this will clear the temp vectors used to store the id/aid/uid/pos for the actions/movements after calling the register() method.


dudantas:
Thanks for the contribution!

PR #596, opened 2019-09-19
fix: purchase prey bonus rerolls in store

westwol:


gerotib:
This probably should be also fixed in dailyreward.lua:
`self:addBonusReroll(reward.rerollCount)` to
`self:setPreyBonusRerolls(self:getPreyBonusRerolls() + reward.rerollCount)`

PR #600, opened 2019-09-20
Update Deepsea_Blood_Crab.xml

cellss:
Its necessary to fix an update conflict

Non-unique loot item "fish"

Cant load loot. data/monster/new/Deepsea_Blood_Crab.xml


cellss:
This change is not needed if you compile with the latest sources.

PR #601, opened 2019-09-20
Update actions.xml

cellss:
wrong tag

[Warning - Event::checkScript] Can not load script: scripts/other/teleport.lua
cannot open data/actions/scripts/other/teleport.lua: No such file or directory
[Warning - Event::checkScript] Can not load script: scripts/other/teleport.lua
cannot open data/actions/scripts/other/teleport.lua: No such file or directory


majestyotbr:
Thanks for the fix.

PR #603, opened 2019-09-21
added Gold converter tag

cellss:
A tag for item: 29020, gold converter for sale on Store...

directed to existing script others/objects/goldConverter.lua


majestyotbr:
Thanks.

PR #605, opened 2019-09-23
Improves the RSA fix to log more information

Costallat:
- Log when the RSA doesn't match the one expected by the server

- Log failed login attempt.

- Add Docker with 'gdb'

- Add local login script to be used to test a local server with client +11 that requires login.php

- Restore commentary with the SQL to create the sample players, as it is useful when not using any website


majestyotbr:
Thanks for your contribution!

PR #611, opened 2019-09-25
Adds spdlog to improve logs, doxygen and ldoc to generate documentation, cmake build type release and debug, update travis and appveyor

Costallat:
- spdlog: log library so we can replace the 'std:out' by calls to that library and also use the log levels 

- doxygen and ldoc: generate the documentation of the code and of the scripts, 

- cmake build type release and debug: there was no difference between then, now the debug will have more warnings but will not fail because of them, and release, that will fail in any warning(less warning enabled)

- travis and appveyor: updated their files to use the new release/debug types, and also created a single zip file on appveyor with the server and all dll's inside it


majestyotbr:
Thanks for your contribution!

PR #617, opened 2019-09-27
market removeMoney

FakeShinoda:
- Market now use the money from bank or inventory or both to create an offer


majestyotbr:
Thanks for your contribution!

PR #612, opened 2019-09-25
Prey Fixes/Adjusts

FakeShinoda:
- Attempt to fix issue #412
- Renew the prey time when use a bonus reroll
- Fix the 2 minutes bug
- Infos in prey mini-window now refresh automatically
- Prey time now decays killing any creature
- Send correct wildcard's prices
- Fixes a bug in preyloot

PR #618, opened 2019-09-27
Fix Prey Damage

FakeShinoda:
- Fix the bug generate by the pull #612 
- Another attempt to fix 0% bonus and 0 stars

Was tested on 12 characters and returned no 0% bonuses, should be fixed.
Steps is not working, need a formula rewrite.


majestyotbr:
Thanks for the fix.

PR #625, opened 2019-09-29
Fix monster respawn

Costallat:
Monster respawn was not working.

Signed-off-by: Renato Foot Guimarães Costallat <costallat@hotmail.com>


majestyotbr:

PR #630, opened 2019-09-30
Added 12.20 items, fixed monster errors because of the item update

oualid6496:
- Added 12.20 items
- Moved some items to the correct place
- Fixed couple small errors
- Added the Omniscient Owl action
- Fixed double item errors because of the new items update


majestyotbr:
Tested, working, no errors on distro. Thanks for your contribution!

PR #631, opened 2019-10-01
Fix bank deposit not adding the amount of money when the bank has some money

Costallat:
Fixes bank deposit, but breaks the feature of using money from the player + bank to by things from the market.


majestyotbr:
Thanks for the fix.

PR #634, opened 2019-10-03
[Market] Removes money from player when it doesn't have enough money on the bank.

Costallat:
If the player doesn't have enough money on bank to create or to buy some offer it will now take into account the money that the player have in the the backpack.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -5601,7 +5601,8 @@ void Game::playerCreateMarketOffer(uint32_t playerId, uint8_t type, uint16_t spr
 	}
 
 	const uint32_t maxOfferCount = g_config.getNumber(ConfigManager::MAX_MARKET_OFFERS_AT_A_TIME_PER_PLAYER);
-	if (maxOfferCount != 0 && IOMarket::getPlayerOfferCount(player->getGUID()) >= maxOfferCount) {
+	if (maxOfferCount != 0 && 
+		IOMarket::getPlayerOfferCount(player->getGUID()) >= maxOfferCount) {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


Costallat:
Can someone test this? 

Just to make sure that it's working ok? 

Tks

gerotib:
Getting fee from player with 0 bank balance while creating offer - works fine. Trying to buy with no money - not possible as expected. Trying to buy item from market with money only on player - works fine. So everything is ok i think.

liontts (approved):

majestyotbr:
Thanks for your contribution!

src/game.cpp
```diff
@@ -5649,18 +5655,37 @@ void Game::playerCreateMarketOffer(uint32_t playerId, uint8_t type, uint16_t spr
...
+		// Have enough money on the bank
+		if(totalPrice <= player->getBankBalance())
+		{
+			player->setBankBalance(player->getBankBalance() - totalPrice);
+		}
+		else
+		{
+			uint64_t remainsPrice = 0;
+			remainsPrice = totalPrice - player->getBankBalance();
+			g_game.removeMoney(player, remainsPrice);
```
cipsucks, line 5687: Bank Balance is not set to 0

dudztroyer:
Actually this should be the other way around... Spend money that the player has in the inventory before spending bank balance... It happens as this when using npc trade/sail... In global it happens like this in market as well... Minor detail tho

cipsucks:
I agree, on global the amount is subtracted from inventory first. I only stumbled across the error because you can dupe money infinitely. Make an offer for 50,000,000 and only have 49,999,999 in your account and 1gp in your inventory, you're only charged 1gp, then cancel the offer and you and have 100kk in your account now.

I don't have time to re-arrange the logic to match global, but I know that is preferred on this project, so rather than make a pull request keeping this logic and adding the one line to empty the bank, I thought I'd just make a comment.

PR #639, opened 2019-10-09
couple of minor fixes

Raphaellb:
I made fixes to the hero of rathleton quest in the past, but I didn't realize I didn't submit this small fix on lava entrance levers.


majestyotbr:
Thanks for the fix.

PR #647, opened 2019-10-12
Update schema.sql

rafaelmmendes:
Wasn't being able to create guilds using MyAAC (SQL error "Field 'residence' doesn't have a default value"). Altering residence default value to '0' solved it.


majestyotbr:
Thanks for the fix.

PR #648, opened 2019-10-13
Update items.xml

rafaelmmendes:
Menacing carnivor's corpse (id="34741") was decaying to value="34742", an item that is not registered in items.xml and has the sprite of a bluish little bird statue, which is not moveable and thus blocking the player's passage. I also corrected the later corpses' stages to their respective proper item id's.


majestyotbr:
Thanks for the fix.

PR #649, opened 2019-10-14
luajit: Fix lua.hpp path detection, especially for vcpkg

Costallat:
This fixes the problem on windows that lua.hpp cannot be found, because vcpkg by default installs luajit, which results in <luajit/lua.hpp>

Signed-off-by: Renato Foot Guimarães Costallat <costallat@hotmail.com>


majestyotbr:
Thanks for the contribution.

PR #652, opened 2019-10-14
Update spawns

lBaah:
Update spawns, since I made use of an outdated file.

Thank you @rickorion


majestyotbr:
Thanks for the fix.

PR #654, opened 2019-10-15
Fix NPC Walking radius

lBaah:
Set NPC default walking radius when there is no one set on NPC XML file.
Set NPC masterpos on spawn. Walking radius will respect the masterpos reference.

Define walking radius on NPC file with tag
> walkradius="3"


FakeShinoda (approved):

FakeShinoda:
LGTM

lBaah:
Thank you for the review.

majestyotbr:
Thanks for the fix.

PR #662, opened 2019-10-16
[TFS]: Define item attribute types in constants

Costallat:
Moved itemAttrTypes to constants to make it more readable.

Also fixed comparing attribute types in functions isIntAttrType, isStrAttrType and isCustomAttrType as it was possible to pass not existing type to it (from Lua).

Signed-off-by: Renato Foot Guimarães Costallat <costallat@hotmail.com>
Co-authored-by: Renato Foot Guimarães Costallat <costallat@hotmail.com>

PR #663, opened 2019-10-17
[TFS] Lua item:decay(decayId) per item

Costallat:
Changed ItemType attribute `decayTo` to ItemAttribute. This attribute is savable to database.
Now it can be customized for every Item.

Previous version of Lua function `item:decay(decayId)` did modify `decayTo` of ItemType. Which did resulted in modifying all items on the server, not just the item on which it was called.


majestyotbr:
Anyone to test this?

gerotib:
Shouldnt [these lines](https://github.com/opentibiabr/OTServBR-Global/pull/663/files#diff-f7ca61cf749eba039da961e8a7be1630L838-R850) be added as separated "if" instead of replacing this `ITEM_ATTRIBUTE_SPECIAL`?

src/item.cpp
```diff
@@ -835,9 +844,9 @@ void Item::serializeAttr(PropWriteStream& propWriteStream) const
 		propWriteStream.write<uint8_t>(getIntAttr(ITEM_ATTRIBUTE_SHOOTRANGE));
 	}
 
-	if (hasAttribute(ITEM_ATTRIBUTE_SPECIAL)) {
-		propWriteStream.write<uint8_t>(ATTR_SPECIAL);
-		propWriteStream.writeString(getStrAttr(ITEM_ATTRIBUTE_SPECIAL));
+	if (hasAttribute(ITEM_ATTRIBUTE_DECAYTO)) {
+		propWriteStream.write<uint8_t>(ATTR_DECAYTO);
+		propWriteStream.write<int32_t>(getIntAttr(ITEM_ATTRIBUTE_DECAYTO));
```
gerotib, line 849: Shouldnt these lines be added as separated "if" instead of replacing this `if (hasAttribute(ITEM_ATTRIBUTE_SPECIAL))`?

src/item.h
```diff
@@ -104,10 +104,11 @@ enum AttrTypes_t {
 	ATTR_ARMOR = 31,
 	ATTR_HITCHANCE = 32,
 	ATTR_SHOOTRANGE = 33,
-	ATTR_SPECIAL = 34,
-	ATTR_IMBUINGSLOTS = 35,
+	ATTR_CUSTOM_ATTRIBUTES = 34,
+	ATTR_DECAYTO = 35,
+	ATTR_IMBUINGSLOTS = 36,
+	ATTR_SPECIAL = 37,
```
gerotib, line 110:
```suggestion
	ATTR_SPECIAL = 37
```

src/item.h
```diff
@@ -504,7 +505,7 @@ class ItemAttributes
 
 	public:
 		static bool isIntAttrType(itemAttrTypes type) {
-			return (type & 0x7FFE13) != 0;
+			return (type & 0xFFFE13) != 0;
```
gerotib, line 508: Should be merged with #662 then it would be:
`return (type & intAttributeTypes) == type;`

gerotib:
I tested this along with #662 (with couple fixes that i pointed in review notes) and everything seems to work fine then

Issue #670, opened 2019-10-20
Market Bugs


FakeShinoda:
Things to fix on Market:

- When you create/buy an offer of TibiaCoins the balance only updates if you re-login
- You can only make an offer  of 100 items/coins at time

PR #671, opened 2019-10-20
Fix Market Coins

FakeShinoda:
These market changes fix issue #670 

- The tibia coins balance displayed on market are now updated automaticaly

- There's a 3000 items limit items for create a offert on market, however the non-stackeble items limit remains 1000

- Fix a bug in Dwarven Legs Quest

PS: Test it before apply. Thanks !


lBaah (approved):

majestyotbr:
Thanks for the fix.

PR #680, opened 2019-10-24
Fix First Dragon Quest

joseamaya1:
Bug fix: When a player enters the boss's room he cannot travel


Raphaellb:
can you be more specific on this problem please? what boss room, and what teleport/tile/lever are you trying to use? what is the purpose of added ZorvoraxTimer = 10223, storage?

MajestyOTBR:
@joseamaya1

majestyotbr:
https://github.com/opentibiabr/OTServBR-Global/blob/develop/data/movements/scripts/quests/first%20dragon/entranceTeleports.lua#L4

majestyotbr:
Thanks for the fix.

Issue #235, opened 2019-07-06
[Cliente 12] Menus não salvando


Preichmann:
**Describe the bug**
Ao relogar todas as janelas que estao abertas(skills,battle,vip) e a barra de hp que fica no topo da tela voltam para a config original, não esta ficando salvo no client 12

1. Logar no cliente 12(branch dev)
2. Abrir abas e mudar a barra de hp para a posicao de baixo(como na print)
3. Relogar com CTRL+G
4. Verificar que todas as abas vem fechadas ao logar e a barra de HP volta para o topo(default)

**Expected behavior**
As configuracoes devem ficar salvas ao relogar/sair do tibia

**Screenshots**
Arrumar as abas assim:
![image](https://user-images.githubusercontent.com/12971964/60759247-619e9780-9ff8-11e9-890a-8bb736b18030.png)
Relogar com CTRL+G
![image](https://user-images.githubusercontent.com/12971964/60759251-69f6d280-9ff8-11e9-9bc5-446e3ed36fec.png)

**Desktop (please complete the following information):**
- Branch Dev
- Client 12
- Hosteado em win 10

Preichmann:
OBS: O problema não parece ser do cliente, pois se o char ficar logado ele volta com a msm config, por exemplo: 
Logo o char e configuro tudo que quero
Saio fechando o cliente no X
Logo novamente no char
as coisas continuam configuradas

Aerwix:
It's server side, we need a special script to save the backpacks/panels, but right now everyone is focused in other things.

Hopefully this feature will be available soon.

gerotib:
Its saving data correctly, but each time you login your character, new folder in client folder "characterdata" is being created, instead of using the one that should be assigned to this character.

vinnyzeramoc:
qual a solução para isso ?

gerotib:
I made it work by changing function that set player id on login from auto incremented playerAutoID++ to this playerAutoID number which was 0x10000000 + guid (with this player gets same id each login, and client load proper files with settings). Then I had to increase this playerAutoID number so getCreatureByID works properly. I dont think this is a great solution, so I didnt created any pr, but it works fine and requires very few code changes. Heres exactly what ive changed:
in player.h:
```
	void setID() final {
			if (id == 0) {
				id = playerAutoID++;
			}
		}
```
to:
```
		void setID() final {
			if (id == 0) {
				if (guid != 0) {
					id = 0x10000000 + guid;
				}
[...]

PR #681, opened 2019-10-24
Fix Save Layout Client11+

hygorrasec:
This update fixes the problem that when the player logout, the layout adjusted by him is not saved.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -48,7 +48,7 @@ extern Imbuements* g_imbuements;
 
 MuteCountMap Player::muteCountMap;
 
-uint32_t Player::playerAutoID = 0x10000000;
+uint32_t Player::playerAutoID = 0x10010000;
 
 Player::Player(ProtocolGame_ptr p) :
 	Creature(), lastPing(OTSYS_TIME()), lastPong(lastPing), inbox(new Inbox(ITEM_INBOX)), client(std::move(p))
--- a/src/player.h
+++ b/src/player.h
@@ -142,7 +142,9 @@ class Player final : public Creature, public Cylinder
 
 		void setID() final {
 			if (id == 0) {
-				id = playerAutoID++;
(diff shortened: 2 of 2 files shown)


miezaF:
Server can be bugged with this solution.
Also it would bug with over than 10.000 players on db.

src/player.cpp
```diff
@@ -48,7 +48,7 @@ extern Imbuements* g_imbuements;
 
 MuteCountMap Player::muteCountMap;
 
-uint32_t Player::playerAutoID = 0x10000000;
+uint32_t Player::playerAutoID = 0x10010000;
```
FakeShinoda, line 51: so you can olny have 10.000 players ?

hygorrasec:
> Server can be bugged with this solution.
> Also it would bug with over than 10.000 players on db.

Hi, I have not found another solution though 10k+ players online is almost impossible. The problem is just this?

FakeShinoda:
In this case "playerAutoID" will be the same forever ?

hygorrasec:
> In this case "playerAutoID" will be the same forever ?

I noticed that with this playerAutoID he was creating new ID's in the /characterdata folder every time the same char entered. So it dont save the layout within the game.

gerotib:
Thats what I commented in #235, and yeah it works, but needs improvements as you guys realised. Hope we can find actually good and bug-free solution ;)

hygorrasec:
> Thats what I commented in #235, and yeah it works, but needs improvements as you guys realised. Hope we can find actually good and bug-free solution ;)

Hey! As I dont find any PR, I did so to try solve this problem. Sorry for not seeing your issue before, this solution I found in a forum, adapted it and it worked.

Raphaellb:
I tested this yesterday with latest build, and it is working.

majestyotbr:
Tested, working. Thanks for the contribution.

PR #688, opened 2019-10-26
Option to Disable/Enable Container Weight

hygorrasec:
I created a variable (**DISABLE_CONTAINER_WEIGHT**) in data/events/player.lua to make it easier to **enable(0)**/**disable(1)** container's movement depending on its weight.

_The example below shows the impossibility of moving a backpack over 1000 oz:_
[][]


majestyotbr:
Thanks for the contribution.

PR #698, opened 2019-10-27
Making item name singular, so that when looked

rafael-adcp:
Making item name singular, so that when looked it doesnt end with "coinss", also addingscript to handle click on tibia coin, so that it can be added into players account'

### DIFF
--- a/data/actions/actions.xml
+++ b/data/actions/actions.xml
@@ -1125,6 +1125,7 @@
 
 	<!-- Market -->
 	<action itemid="14405" function="market" />
+	<action itemid="24774" script="others/tibiacoin.lua"/> 
 
 	<!-- Dolls -->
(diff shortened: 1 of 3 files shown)


majestyotbr:
Custom action.

lBaah:
It is "coins" because you can only trade stacks of 25 of them in the market.

rafael-adcp:
hmm ok fair, agree with custom action, so 2 questions then, how are coins supposed to get to a player?
also if you use a gm and do /i tibia coin, 25 and look it will show as "tibia coinss" which is wrong imo

majestyotbr:
Don't worry, will be fixed.

lBaah:
This is not an obtainable item. Coins should be sent through AAC.
That item is necessary only to be displayed on market and Cyclopedia Items.

PR #679, opened 2019-10-23
Add Server Save configuration to the config.lua file

Costallat:
Add configuration to enable/disable server save notify

Signed-off-by: Renato Foot <rcostallat@daitangroup.com>

PR #699, opened 2019-10-27
Fix next reward countdown time

### LINKED ISSUE #257: Daily reward
Boa noite,

Fiz alguns testes, ele deixa voce pegar o premio do dia, porem nao inicia a contagem para a proxima que você pode pegar e com isso não é possivel pegar o segundo reward.

gerotib:
It fixes countdown time to next server save (24 hours from previous one).
I think it was set like this for testing purposes.
Along with #679 it fixes #257.
daily


majestyotbr:

PR #705, opened 2019-10-29
[Feature] docker-compose for development environment

leoddias:
With the existing scripts I created a new way to develop and start the necessary services in Docker using docker-compose.


Costallat:
Tested and its working now! All services are started without errors!

Thank you for your cooperation!

majestyotbr:
Thanks for the contribution.

PR #708, opened 2019-10-30
[TFS] Show zero in keys numbers

Costallat:
Now keys will display (Key:0021) instead of (Key:21);


leoddias:
WARNING: You made two changes in this PR.

PR #716, opened 2019-11-01
Fix: Rainbow Necklace and Moon Mirror

hygorrasec:
Included Rainbow Necklace attributes and change Lit Moon Mirror to Moon Mirror.


dudantas:
Thanks for your contribution!

PR #717, opened 2019-11-01
Fix: Mysterious Scroll

hygorrasec:
The itemID being used was from another quest (https://www.tibiawiki.com.br/wiki/Translation_Scroll). I changed it to a similar sprite that was dont use.


majestyotbr:
Thanks for the fix.

PR #718, opened 2019-11-01
Fix: Boat kick system

hygorrasec:
I removed commented of boat kick system and made some coordinate corrections for some.


majestyotbr:

PR #719, opened 2019-11-02
Storage management talkaction

lBaah:
Talkactions to help developers working with storage

/set playername, storage, value
/get playername, storage


dudantas:
Thanks for your contribution!

PR #720, opened 2019-11-02
Fixed: NPC Charos travel

hygorrasec:
Fixed like https://www.tibiawiki.com.br/Charos
The trip is not need the quest, at first it is so.


majestyotbr:

PR #727, opened 2019-11-04
Allow server save time customization

Costallat:
Several improvements to allow customization and gave a better name to the variables


majestyotbr:
Tested and it's working. Thanks for your contribution!

PR #734, opened 2019-11-05
Fix party loot in Loot channel

gerotib:
`MESSAGE_LOOT` automatically switches to "Loot" channel if this channel is open, else it broadcasts in "Server Log". Without this change party members always get loot info in "Server Log", even when "Loot" channel is open.
loot


lBaah (approved):

majestyotbr:
LGTM

PR #738, opened 2019-11-07
Fix client magic level display while decreased with ConditionAttributes

gerotib:
fix based on #374 

Example: your mlvl is decreased by silencer, in client it still show your basic magic level instead of decreased value.

With this change mlvl is updated correctly every time you get or lose condition that decrease it.


majestyotbr:
Thanks for the fix.

PR #742, opened 2019-11-07
Update damage

lBaah:
remove duplicated halved damage

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -530,25 +530,9 @@ void Combat::CombatManaFunc(Creature* caster, Creature* target, const CombatPara
 {
 	assert(data);
 	CombatDamage damage = *data;
-	if ((damage.primary.value < 0 || damage.secondary.value < 0)) {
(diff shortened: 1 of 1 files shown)


enzerah:
don't look good to me bruh

lBaah:
Great review @EnzzoCaaue. Lot's of information and data provided as never seen before.
Certainly this is going to change everything.

enzerah:
yeah but this is to set crit for non black skull players and black skull players, since theyr damage are different

gerotib:
As I am testing it doesnt matter at all (if were talking about damage on players with mana shield). With or w/o this commit regular and critical damage on player with & without BS works correct. Looks like this critical applying only matter in CombatHealthFunc and are affecting mana the same way.
Note that I am not using 100% clear repo code now (and cant really do at the moment), but I dont think that I made any change that could affect this.

lBaah:
You're totally correct @gerotib. The damage/critical is applied and calculated in health function.
This commit works properly for you?

gerotib:
Yeah everything fine

PR #748, opened 2019-11-09
Add server save configuration

Costallat:
Add the server save configuration to the config file.


MajestyOTBR:
We will maintain the default serversave, without custom things.

Thanks.

dudantas:
your fix is only ";"? are you kidding me?

PR #757, opened 2019-11-11
Update dailyreward.lua

Raphaellb:
Currently when the player gets the re-roll rewards from daily rewards the script is setting that count to 2, instead of getting the current amount the player has, and adding 2 to that amount. this fixes that problem.


lBaah (changes requested):
Uncomment line 512 ExpBoost related

PR #760, opened 2019-11-12
Small Fix: Container/Item Weight

hygorrasec:
- Small fix in player.lua about blocking container movement when overweight;
- I also removed this lock from tile.cpp (the function did the same thing);
- Optimization of container.cpp


gerotib (approved):

majestyotbr:
Tested, working.

PR #763, opened 2019-11-13
Daily reward updates

gerotib:
1. `DAILY_REWARD_TYPE_RUNE_POTION` changed to `DAILY_REWARD_TYPE_ITEM`
2. `DAILY_REWARD_TYPE_STORAGE` commented as template for storage reward
3. rune/potion rewards like rl
4. 5th reward = second wildcard instead of storage 
5. 6th reward = training weapons
6. added `itemCharges` used for training weapon charges
7. commented not so useful print
8. uncommented xp boost applying section, and commented warning window that shows without checking anything
9. streak lost message wont appear at first character login

Also applied fix from #757 so it could be closed if not merged before.


majestyotbr:
Tested, working.

PR #768, opened 2019-11-18
Fix player capacity verify before buy item in Store

marcosogsantos:


majestyotbr:

PR #774, opened 2019-11-23
Fix walk delay of creatures

marcosogsantos:
Creatures as White Deer that must run of players have a biggest delay to walk, making too easy catch them. Since I didn't find any monster related walking interval, I assumed that the problem was in creatures "thinking" interval and changed it from 1000 to 200. I haven't studied how much this affects in the server performance, but fixed the problem and all looks fine.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -72,7 +72,7 @@ class Item;
 class Tile;
 
 static constexpr int32_t EVENT_CREATURECOUNT = 10;
-static constexpr int32_t EVENT_CREATURE_THINK_INTERVAL = 1000;
+static constexpr int32_t EVENT_CREATURE_THINK_INTERVAL = 200;
 static constexpr int32_t EVENT_CHECK_CREATURE_INTERVAL = (EVENT_CREATURE_THINK_INTERVAL / EVENT_CREATURECOUNT);
 
 class FrozenPathingConditionCall


Lycanzito:
> Creatures as White Deer that must run of players have a biggest delay to walk, making too easy catch them. Since I didn't find any monster related walking interval, I assumed that the problem was in creatures "thinking" interval and changed it from 1000 to 200. I haven't studied how much this affects in the server performance, but fixed the problem and all looks fine.

Im Going to  test it, did you find something about the "radius" distance? 
I mean, the distance from the player that the monster run from ?

Lycanzito:
well.. with this amount the monsters are running too too too fast, I will try with 800

Lycanzito:
> well.. with this amount the monsters are running too too too fast, I will try with 800

Tested with 800, but the creatures still look strange.. they "go back" , walk strange.. etc..
Dont think that this pull request is necessary..

marcosogsantos:
The problem its not the range, but the walk and stop process of monsters that should run from player. That 1000ms delay on "thinking" cause a not fluid escape.
200ms was my first try and run well, but maybe a bigger number between 300 and 500 can fix it too...

dudantas:
The problem is that 1000ms is a default value, as are the players. Changing this value doesn't solve the problem, it just creates others. We should place a "walkinterval" in the script of monsters that have to have a smaller interval than the others.

marcosogsantos:
You are right, my bad. So, I find another away to "fix" the walk and stop - or another hack. 
This [line](https://github.com/opentibiabr/otservbr-global/blob/11d2a0001ed2534171d4ebae0c59c35f1ad9122e/src/creature.cpp#L926) tell to monster add one sqm of distance into their list of walk, so if do you double it, like
`listWalkDir.push_front(dir);`
`listWalkDir.push_front(dir);`
the monster will walk 2 times to one direction, and avoid the pause to walk. All above "isFleeing()" or "maxTargetDist" conditionals
@dudantas @Lycanzito

PR #791, opened 2019-12-21
Changed monster speed

Art4k:
According to [TibiaWiki](https://tibia.fandom.com/wiki/Bat) Bat speed should be 100.


majestyotbr:
All monsters of datapack need a revision. Anyway, thanks for your contribution.

PR #800, opened 2019-12-30
Fix market buy offer with player offline

Costallat:
Fix player with an open buying offer not receiving the item when it is offline.


Lycanzito:
Erro fixado, obrigado <3
!

PR #808, opened 2020-01-08
fix creature walking issues

lBaah:


mattyx14 (approved):

Lycanzito:
any relations with #749  ?

lBaah:
this resolves #749

Lycanzito:
> this resolves #749

God bless you, tested it and it worked!! :D but one of the problems is still "avaliable"

The monster is still getting only 10 SQMS from the player.

image

lBaah:
What you ask for is not related with this PR.
You can just increare it here https://github.com/opentibiabr/otservbr-global/blob/9732a56545584f0a0346796165fb5072c56c0189/src/monster.cpp#L90
Therefore you'll might have new issues.

Lycanzito:
> What you ask for is not related with this PR.
> You can just increare it here
> 
> https://github.com/opentibiabr/otservbr-global/blob/9732a56545584f0a0346796165fb5072c56c0189/src/monster.cpp#L90
> 
> 
> Therefore you'll might have new issues.



> What you ask for is not related with this PR.
> You can just increare it here
> 
> https://github.com/opentibiabr/otservbr-global/blob/9732a56545584f0a0346796165fb5072c56c0189/src/monster.cpp#L90
> 
> 
> Therefore you'll might have new issues.

Changed to 22,22 , instead of 10,10, and the creature just stop, after running, the 22 sqms and dont move again..

u.u

lBaah:
So move it to 

`return Creature::canSee(getPosition(), pos, std::numeric_limits<int>::max(), std::numeric_limits<int>::max());`

 and good luck.

Lycanzito:
> So move it to
> 
> `return Creature::canSee(getPosition(), pos, std::numeric_limits<int>::max(), std::numeric_limits<int>::max());`
> 
> and good luck.

Can you be more specific? please? ^.^

PR #809, opened 2020-01-08
feat: colored loot message

lBaah:
doesn't work on older client
missing creaturescript at scripts/creaturescripts yet

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -204,7 +204,7 @@ std::ostringstream& Container::getContentDescription(std::ostringstream& os) con
 			os << ", ";
 		}
 
-		os << item->getNameDescription();
+		os << "{" << item->getClientID() << "|" << item->getNameDescription() << "}";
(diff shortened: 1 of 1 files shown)


gpedro:
LGTM. Should check the version before sending, no?

Lycanzito:
it should be colored, right? '-' 
image

mattyx14:
he say it: missing creaturescript at scripts/creaturescripts yet

lBaah:
This is meant to use with OTBR not custom servers @Lycanzito .

@gpedro we should not support outdated client. This allows us to focus on the development of protocol 12 features.
Cipsoft has changed the way few things are rendered in the screen (12.15+) thus making outdated not compatible anymore, if we update the protocol more than 12.00. So we drop the support.

lBaah:
I'll close this since it is going to be a 1.0 feature.

PR #810, opened 2020-01-08
fix client crash when trading lot of items with npcs

lBaah:
this code send a lot of useless message to the client, basically with the same info, therefore, freezing the client.
we need to send it just one time.


mattyx14 (approved):

cmaleister:
In fact, the customer no longer crashes, but anyway he stops working when the sale is high.
I tried to sell 62 magic plate armor in different backpacks.

Lycanzito:
here, the problem remains the same.
image

lBaah:
I'll close this since its not for develop.

PR #811, opened 2020-01-08
drop outdated client support

lBaah:
this will drop outdated client support and allow into the server 12.00 protocol only


mattyx14:
like it

gpedro:
nice idea!! can you check lua files? Search by 
> getClient().version

lBaah:
I'm going to work with it on the 1.0.

PR #823, opened 2020-01-24
Update creature.lua

EleanorRigby1:
It is not my proposal to remove it, the system is good if it works, I don't know much about programming, so I didn't fix it. But that corrected the error.

With that code, when you have White Skull and let the monsters attack you, you give it to exit and re-enter as in 20 seconds you will be kicked with white skull.

### DIFF
--- a/data/events/scripts/creature.lua
+++ b/data/events/scripts/creature.lua
@@ -74,6 +74,29 @@ function Creature:onTargetCombat(target)
 			__picif[target.uid] = {}
 		end
 	end
+	
+	if target:isPlayer() then
+		if self:isMonster() then
+			local protectionStorage = target:getStorageValue(Storage.combatProtectionStorage)
+
+			if target:getIp() == 0 then -- If player is disconnected, monster shall ignore to attack the player
+			    if target:isPzLocked() then return true end
+				if protectionStorage <= 0 then
+					addEvent(removeCombatProtection, 30 * 1000, target.uid)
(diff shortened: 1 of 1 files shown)


data/events/scripts/creature.lua
```diff
@@ -75,29 +75,6 @@ function Creature:onTargetCombat(target)
 		end
 	end
 
-	if target:isPlayer() then
-		if self:isMonster() then
-			local protectionStorage = target:getStorageValue(Storage.combatProtectionStorage)
-
-			if target:getIp() == 0 then -- If player is disconnected, monster shall ignore to attack the player
-			    if target:isPzLocked() then end
```
FakeShinoda, line 83: if target:isPzLocked() then return true end
this should fix it

mattyx14: no require remove all function only require add condition skull or battle

lBaah: isPzLocked() checks if the player is in fight (pz block), no matter if skulled or not, so it might work

lBaah (changes requested):
Please keep the function and change the line 83: `if target:isPzLocked() then end` to `if target:isPzLocked() then return true end`

EleanorRigby1 (commented):
ty

lBaah (approved):

lBaah:
Thank you for the changes you made.

Issue #340, opened 2019-08-01
[Items] Store tapestry can be wrapped inside the wall


oualid6496:
When you buy a tapestry in the store and place it on the wall you can wrap it again and then you'll have a decoration kit inside a wall.

ghost:
I don't think that the tapestries hanged sprite should be wrappable at all, to prevent this issue

oualid6496:
exactlly but it has the same item id as the one that you can pickup so im not sure how to change this

ghost:
true, my bad, the wrappable system need a review

rilex037:
So is it like a toggle? Is there no lua function to get sprite frame from item?

oualid6496:
Not that I know of I have no clue to where this is being handled

dudantas:
This is handled in items.xml

PR #833, opened 2020-02-04
Fix for status bar not showing correct amount of items

Costallat:
- Status bar doesn't show full amount of items

### DIFF
--- a/docker/server-only/login.py
+++ b/docker/server-only/login.py
@@ -50,12 +50,12 @@ def do_login(data):
         };
         print(session)
 
-
-        sql_select_Query = "SELECT name, level, sex, vocation, looktype, lookhead, lookbody, looklegs, lookfeet, lookaddons, lastlogin from players where id = '" + str(account_id) + "'"
+        sql_select_Query = "SELECT name, level, sex, vocation, looktype, lookhead, lookbody, looklegs, lookfeet, lookaddons, lastlogin from players where account_id = '" + str(account_id) + "'"
 
         cursor = connection.cursor()
         cursor.execute(sql_select_Query)
         records = cursor.fetchall()
+        print(records)
 
         players = []
         vocations = ['None', 'Sorcerer', 'Druid', 'Paladin', 'Knight']
@@ -76,8 +76,7 @@ def do_login(data):
                     'tutorial': False, # True will make the client crash as it will try to auto connect
                     'ishidden': False,
                     'istournamentparticipant': False,
-                    'remainingdailytournamentplaytime': 0
-                    
+                    'remainingdailytournamentplaytime': 0            
             }
             players.append(player)
         print(players)
@@ -86,7 +85,7 @@ def do_login(data):
             'worlds': [
                 {
                     'id': 0,
-                    'name': 'Open-TibiaBR',
+                    'name': 'OTServBR-Global',
                     'externaladdressprotected': '127.0.0.1',
                     'externalportprotected': 7172,
                     'externaladdressunprotected': '127.0.0.1',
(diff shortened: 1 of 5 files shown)


gerotib:
@Costallat I tested and sending items to depot works correctly, but I am unable to **unwrap** items, thats because of `|| !item->isWrapable() `, but when I change it to `|| item->isWrapable() `, then its not possible to **wrap** an item. I wonder if that check is even needed, because it works fine w/o this part.

Lycanzito:
> @Costallat I tested and sending items to depot works correctly, but I am unable to **unwrap** items, thats because of `|| !item->isWrapable() `, but when I change it to `|| item->isWrapable() `, then its not possible to **wrap** an item. I wonder if that check is even needed, because it works fine w/o this part.


When I try to unwrap a item:
image
dont works

when i "use"
it works
image
image


and the item cant be "wrap" again 
image


also itens can be though outside the house
image


But now with this changes, only is possible to "unwrap" only inside the house:
unwrap shows:
image

use shows:
image

Costallat:
@gerotib @Lycanzito 

That's because I am trying to block to unwrap things on walls, to fix the issue #340 , so the change that I made didn't work.

Costallat:
Test again

Lycanzito:
> Test again

I will test it, in a couple of minutes, i give u the feedback

Costallat:
Tks,

So I will keep only the fix of the status bar and the store items sent back to the store when the player leaves the house.

Costallat:
Someone can do a final test with those two fixes? 

Tks,

Lycanzito:
> Someone can do a final test with those two fixes?
> 
> Tks,

Tested.. The itens are going to the inbox, right?
image

src/game.cpp
```diff
@@ -2423,7 +2423,9 @@ void Game::playerWrapableItem(uint32_t playerId, const Position& pos, uint8_t st
 		return;
 	}
 
-	if (!item || item->getClientID() != spriteId || item->hasAttribute(ITEM_ATTRIBUTE_UNIQUEID) || item->isWrapable()) {
+	if (!item || item->getClientID() != spriteId 
+			  || !item->hasAttribute(ITEM_ATTRIBUTE_UNIQUEID) 
+			  || !item->isWrapable()) {
```
gerotib: I removed this part and then its possible to wrap and unwrap items as it should. Do we really need that check?

Costallat: Yes, you need to check if the item is wrapable, ou you will be able to use wrap / unwrap on all items, the problem is that method is not working properly, for some reason  it is returning false when the item can be wrap/unwrap

Lycanzito:
Well.. But with these changes, we lost the possibility to wrap the item back, right?

Costallat:
> Well.. But with these changes, we lost the possibility to wrap the item back, right?

Yes, because it only works because the code is wrong, so we need to fix the code and this is the first step

Costallat:
Test now

Lycanzito:
> Test now

What should these changes make?

Costallat:
- Wrap/unwrap using 'wrap'
-  Drops wrap/unwrap using 'use'

Lycanzito:
> * Wrap/unwrap using 'wrap'
> * Drops wrap/unwrap using 'use'

ok, will test it, and give u the feedback soon

Lycanzito:
maybe i've made something wrong.. but..

image

docker/server-only/login.py
```diff
@@ -87,9 +87,9 @@ def do_login(data):
                 {
                     'id': 0,
                     'name': 'Open-TibiaBR',
-                    'externaladdressprotected': '127.0.0.1',
+                    'externaladdressprotected': '149.56.124.82',
```
gpedro: whoops!

majestyotbr:
Tested, working. Thanks for the fix!

PR #843, opened 2020-02-13
Minor corrections based on other bases

vagnerkuntz:
Reduce cold caches by using vector instead of list for checkCreatures and checkDecay
Fix monster behaviour after despawning
Fix some potential luascript functions crashes
Update on the wrap / unwrap function
Fix spellbook memory leak

Credits:
- SaiyansKing
- TFS


mattyx14:
i test this changes and make crashes server on creaturewalk

vagnerkuntz:
Explains better, because I'm using and so far without problems

PR #852, opened 2020-02-17
Fix Deposit

FakeShinoda:
- Fixed an issue when you trying to deposit an amount higher then you have in hands.


majestyotbr:

Issue #830, opened 2020-01-30
Doors Falcons Not Work


manifex30890:
the door on the falcon doesn't want to work properly because you can't go through it after killing the boss

dudantas:
@manifex30890 you need inform the positions.

dudantas:
Closed because details are missing from the information.

PR #859, opened 2020-02-19
Fix: #781 and #830

dudantas:
Fixed problems mentioned in issue #781 and #830.
Now the door script will check if there is a place to push, when using the door with a player inside, if there is no free SQM, it returns the message "Sorry, not possible."

Thanks for @lBaah for developing the function of checking tiles.


dudantas:
@manifex30890 and @Lycanzito test this PR.

Lycanzito:
> @manifex30890 and @Lycanzito test this PR.

thanks !
I've fixed it doing other things, will make these changes, seens better! =)
then my solution xD

dudantas:
When you test it, return it so we can pass it.

dudantas:
When you test it, return it so we can merge.

majestyotbr:
Thanks.

PR #862, opened 2020-02-20
count-premium-days

lucasgiovannibr:


Lycanzito:
With this change, it shows the number of days in the client login, is that it? sorry if i'm wrong

vagnerkuntz:
> Com essa alteração, mostra o número de dias no login do cliente, é isso? desculpe se estou errado

Não está alteração faz descontar os dias de premium account.
exemplo: hoje dia 19 tenho 30 dias de premium
amanhã no dia 20 terei 29 dias de premium

gpedro:
thanks for contribution <3

should be working, but it is not focused on our project. if you look about the entire world, we have a lot of different people from different countries with different time zones ... so if we put that job on the database probably will appear another infinite question about that and we will be forced to remove this to avoid dumb questions with "stranges" related to premium days.

PR #868, opened 2020-02-21
New door system and reorganization of the folder libs.

dudantas:
Now everything will be organized in just one table, containing each type of door (key locked / key unlocked / custom door / level door / quest door).
Easy to understand, only with openDoor and closedDoor to configure, just register and use, it is not necessary to add the door ID in action.xml and movements.xml, revscriptsys already takes the ids from the table and automatically registers it in onUse..

It fixes the task 3 of issue #644

Global.lua synchronized with TFS, removed tables from doors (now the table is in data / libs / tables / door.lua).
GetTibianTime function replaced by getFormattedWorldTime (synchronizing with TFS).

Credits to me, @gpedro, @MajestyOTBR and @lBaah.


mattyx14:
Is a More simple

Issue #876, opened 2020-02-25
Bug in key locked doors


VitorNM:
## CAUTION: OUR SERVER IS UPDATED DAILY. MAKE SURE YOU ARE REPORTING SOMETHING ON THE LATEST VERSION OF THE SERVER.

**Describe the bug**
The doors that are key locked are impossible to unlock with the correctly key.

**To Reproduce**
1. Find a key locked door (i'll use the bear room door quest)
2. When you try to open the door without using the key, you receive the message that the door is locked.
3. When you try to open the door using the key to unlock it, you receive the message "You cannot use this object." and the door still locked and closed.

**Expected behavior**
When you try to open a key locked door without using a key, you need to receive the message that it is locked, at this point is ok because at this moment we receive this message correctly, but when you try to unlock with the key, the door needs to be unlocked and you can open and close the door without problems and when you use the key in the door again, it needs to be locked and closed.

**Screenshots**
I cant post a screenshot right now.

PR #877, opened 2020-02-25
Fix issue #876 , bug in the open key doors

dudantas:
Fix for the #876 
Problem is only incorrectly key ids (was in clientid).
The variable for the key ids has been moved to the doors table.


VitorNM:
Tested. The issue are correctly fixed. Thanks for the contribution Edu :)

Issue #585, opened 2019-09-17
The first dragon quest


ivansho92:
In the quest, The first dragon, when you operated the level, don't send with the boss, but with the "GOD" i was use this command /goto The first dragon and surprise, just is a npc, you never fight with "Fallen Challenger" and all the chest are emptys.

majestyotbr:
Repeated, #427

ivansho92:
ohhh thank you so much bro

Issue #427, opened 2019-08-18
First Dragon Quest


Raphaellb:
- [x] 1 - after starting the quest, and killing dragons for the Power Boss (Tazhadur). I am getting an error on quest log:

> `Lua Script Error: [Event Interface]`
> data/events/scripts/player.lua:Player@onRequestQuestLine
> data/lib/core/quests.lua:3163: bad argument #2 to 'format' (no value)
> stack traceback:
>         [C]: in ?
>         [C]: in function 'format'
>         data/lib/core/quests.lua:3163: in function 'value'
>         data/lib/core/quests.lua:3387: in function <data/lib/core/quests.lua:3385>
>         (...tail calls...)
>         data/lib/core/quests.lua:3617: in function 'sendQuestLine'
>         data/events/scripts/player.lua:833: in function `<data/events/scripts/player.lua:832>`

- [x] 2 - All movement teleports into the bosses Zorvorax/Gelidrazah the Frozen/Kalyassa/Tazhadur are open to all players after starting the quest, and receiving start storages 10195.
The movement script in movement/quests/first dragon/entranceTeleports.lua is not checking for the specific bosses storages before allowing player into boss fight. Because of this all the player needs to do is start the quest with tamoril, and they are allowed to kill all bosses.

- [x] 3 - Life Boss Zorvorax secrets. player is getting questlog update on the ab'dendriel part by stepping on this position:
> `You see grass.
> Item ID: 9050
> Position: 32691, 31724, 7.
- The correct location is
> 00:00 You see grass.
> Item ID: 9043
> Position: 32697, 31718, 7`

- [x] 4 - Final lever not working: 

> 21:22 You see a lever.

> Item ID: 9826, Action ID: 24888
> Position: 33583, 30992, 14

Raphaellb:
All movement teleports into the bosses Zorvorax/Gelidrazah the Frozen/Kalyassa/Tazhadur are open to all players after starting the quest, and receiving start storages 10195 = 1. the movement script in movement/quests/first dragon/entranceTeleports.lua is not checking for the specific bosses storages before allowing player into boss fight. because of this all the player needs to do is start the quest with tamoril, and they are allowed to kill all bosses.

Raphaellb:
Life Boss Zorvorax secrets. player is getting questlog update on the ab'dendriel part by stepping on this position 00:00 You see grass.
Item ID: 9050
Position: 32691, 31724, 7.
the correct location is 
[...]

PR #878, opened 2020-02-25
Full revision on the first dragon quest and conversion for revscriptsys.

dudantas:
Fix for the #427 

> Added the action / unique loading system per Lua:

There is a major update on the map and it is not possible to add action or unique at the moment, so the script is necessary.

> Revision on the first dragon quest:

> Bugs fixeds:

It was not possible to enter the boss TPS after finishing tasks.
Task Life for boss Zorvorax was not functional, the movements were missing to collect the accesses.

Task Power for boss Tazhadur was not functional, corrected the access teleport.

Treasure for boss Kalyassa was not functional, it was not possible to enter the TP after collecting the 5 rewards and the achievement was given after collecting the first, and not after collecting the 20 (corrected). The achievement did not exist either, it was placed.

Fixed access to the Dragon Lair with the Scale of Gelidrazah, Horn of Kalyassa, Bones of Zorvorax and Tooth of Tazhadur items when using them in front of the gate.

Fixed lever for the Fallen Challenger. Tested and working.

It fixes the #427 #155 #585

Thanks @Raphaellb for the reports.


majestyotbr:
Thanks!

PR #903, opened 2020-03-02
Save attributes and characters in a certain period of time

omarcopires:
The script originally existed, but for some reason I don't know, it was removed.
What currently happens is that if the server crashes or is closed without saving it will return to the last server save, as there is no script that saves every time, this can be configured for every 30 minutes or one hour, according to the particularity of each user, but I recommend that it is every 30 minutes.

### DIFF
--- a/data/globalevents/globalevents.xml
+++ b/data/globalevents/globalevents.xml
@@ -53,6 +53,7 @@
 	<globalevent type="startup" name="ServerStartup" script="startup.lua" />
 	<globalevent type="record" name="PlayerRecord" script="record.lua" />
 	<globalevent name="ServerSave" time="05:55:00" script="serversave.lua" />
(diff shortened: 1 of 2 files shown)


mattyx14 (commented):
``` LUA
function onThink(interval)
	saveServer()
	return true
end
```

dudantas:
> ```lua
> function onThink(interval)
> 	saveServer()
> 	return true
> end
> ```

It's better this way.
The clean map is unnecessary, better to leave the clean only for the global save.

majestyotbr:
Do the requested changes, and the PR can be merged later.

dudantas:
If you want to make the suggested changes, let us know that we will reopen the discussion and merge the pr, thanks anyway.

PR #905, opened 2020-03-03
Fix small, winged backpack 240 slots > 24 slots.

EleanorRigby1:


majestyotbr:
It's not the right way to update a file like items.xml, can't see the diff. You can use git for Linux or github desktop for Windows.

Anyway, thanks. We will fix the slots for winged backpack.

PR #914, opened 2020-03-06
Fix Double Crit

hyresu:
The functions doCombat are calculating the probabilities of a crit (in area or target, both are doing it)
But the function doCombatHealth calls CombatHealthFunc, which calculates crit chance for each target

For example I got this in my server:
22:34 A training monk loses 1012 hitpoints due to an attack by Roshan.
22:34 A training monk loses 1012 hitpoints due to an attack by Roshan. <- Normal hit

22:34 A training monk loses 2277 hitpoints due to an attack by Roshan. <- Aoe calculated crit (doCombat) * Single calculated crit (combatHealthFunc)
22:34 A training monk loses 1518 hitpoints due to an attack by Roshan. <- Aoe calculated crit (doCombat)

With this fix the crit will only be calculated in doCombatHealth and doCombatMana (1 time)
This will also help making crit as normal the doAreaCombatHealth function and doAreaCombatMana


majestyotbr:
Thanks for the fix!

PR #915, opened 2020-03-06
Fix Unique Loot & Gold Pouch

### LINKED ISSUE #902: Question : Reward System
Hello, everyone.

I've a question about the 'unique item' from the reward system.

<item id="5903" chance="100000" unique="1" />				

with this tag, I should receive just 1 f hat, from each Ferumbras, right?

but this is happening:

1st player:
image

2nd player:
image

### LINKED ISSUE #916: Gold Pouch
(Store Problem) ~
Is possible to buy more than 1 for each character;

FakeShinoda:
- This should fix the unique loot drop from bosses (only one item per boss)
- Fix gold pouch beeing possible to buy more than 1 for each character (thanks to @Lycanzito for the fix)

Closes #902, #916


majestyotbr:
Thanks for the fix.

PR #920, opened 2020-03-09
Fix Twist of Fate and Unfair Fight

### LINKED ISSUE #889: Twist of Fate "activated"
Atualmente : A twist of fate só está sendo ativada, se o último hit for de um player.

Como deveria ser: Deve levar em consideração que, ao receber ao menos, X% de dano de um player, no X período de tempo, ativar a mesma.

hyresu:
Right now, the twist of fate is only being calculated when you are last hitted by a player, when in global is :
"When you die, at least 5% of the damage dealt within the last 5 minutes, or the last fatal strike must have been made by a character or its summons."

And the damage map is being calculated for the pz locked time, when in global it is for 5 minutes.

Additionally, the actual default value for unfairFightReduction is 50, making ALWAYS half the total amount of exp loss, even when no players present.

With this fix, the ToF will work as it should and the players will lose the totalexp, not half of it as default.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -2017,27 +2017,31 @@ void Player::death(Creature* lastHitCreature)
 	loginPosition = town->getTemplePosition();
 
 	if (skillLoss) {
-		uint8_t unfairFightReduction = 50;
-		bool lastHitPlayer = Player::lastHitIsPlayer(lastHitCreature);
-
-		if (lastHitPlayer) {
+		uint8_t unfairFightReduction = 100;
+		int playerDmg = 0;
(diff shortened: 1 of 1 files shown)


Lycanzito:
100%;
tested.
working exactly like the global.
;-)

Lycanzito:
#889
#712
podem ser fechadas, acredito que não seja do interesse do repositório as formulas de enforced etc.

vou fecha-las

Lycanzito:
thx for the contribution!

majestyotbr:
Thanks for your contribution!

PR #924, opened 2020-03-10
Update description column in guilds table

Znote:
This is not a standard TFS 1.x column, and breaks inherent guild creation. (on etc Znote AAC). 
The solution is to simply give it a blank value if not set. 
Adding a guild description for a newly created guild should not be a requirement. 

https://github.com/Znote/ZnoteAAC/issues/385

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -270,7 +270,7 @@ CREATE TABLE IF NOT EXISTS `guilds` (
   `creationdata` int(11) NOT NULL,
   `motd` varchar(255) NOT NULL DEFAULT '',
   `residence` int(11) NOT NULL DEFAULT '0',
-  `description` text NOT NULL,
+  `description` text NOT NULL DEFAULT '',
   `balance` bigint(20) UNSIGNED NOT NULL DEFAULT '0',
   CONSTRAINT `guilds_pk` PRIMARY KEY (`id`),
   CONSTRAINT `guilds_name_unique` UNIQUE (`name`),


mattyx14 (approved):

majestyotbr:
Thanks.

slawkens:
This is wrong. Some MySQL versions doesn't support default value in TEXT field. I had this problem too. The solution would be to change to VARCHAR.

Ref: https://stackoverflow.com/a/16142246

Znote:
@slawkens Which size would be appropriate? I can create a new pull request.

slawkens:
as for a varchar starting from MySQL version 5.0.3 (which is pretty standard now) you can use any value between 0-65,535.

For guild description I would go with around 5000-10000 but I'm don't know where this field is used in this distro. Is there some kind of command to set guild description?

Znote:
Just did some scans over the distro, not seeing this used at all. My guess is that its clutter from some Gesior AAC. 
5000 sounds like overkill though.

Found it:
https://github.com/gesior/Gesior2012/blob/80c71ca714ddba4b5abcaf80c114eb73e0f335ec/install.php#L227
https://github.com/gesior/Gesior2012/blob/80c71ca714ddba4b5abcaf80c114eb73e0f335ec/config/config.php#L60

This is AAC clutter, and by default limited to max 2000. Which I consider plenty actually, most guilds with description seems to be below 500. There is also a 6 lines limit by default. 

(Checked these 2):
https://kivera-global.net/?subtopic=guilds
https://aurera-global.com/?subtopic=guilds

slawkens:
Yeah I found it too. Its in MyAAC too.

So I would suggest removing it completely from this distro as it's not required.

Also I noticed following fields in accounts table are not needed too: premium_points, create_date, create_ip, birth_date

Znote:
I think the OTBR teams purposely leaves it in to make it easier dealing with Gesior? The schema file looks like an phpmyadmin export routine.

slawkens:
Yes, but Gesior will install it anyway, and MyAAC too. So it's not needed here I think.

Let's remove it completely? And some other fields I mentioned, too?

Znote:
If we want to clean the schema from AAC clutter, there are lots of things that needs to be changed. Not just this column. So I will just make a correctional PR, but perhaps they are up to it if you submit a proper db clean PR.

PR #927, opened 2020-03-14
Update description column in guilds table

Znote:
Following #924 @slawkens mentioned that the TEXT column doesn't properly support default value in some mysql windows configurations. 

So as a compatibility workaround, we identified that this column is only used by Gesior, and so the column was changed to fit those parameters better in the form of the varchar column type. Which also properly supports default values.

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -270,7 +270,7 @@ CREATE TABLE IF NOT EXISTS `guilds` (
   `creationdata` int(11) NOT NULL,
   `motd` varchar(255) NOT NULL DEFAULT '',
   `residence` int(11) NOT NULL DEFAULT '0',
-  `description` text NOT NULL DEFAULT '',
+  `description` VARCHAR(2000) NOT NULL DEFAULT '',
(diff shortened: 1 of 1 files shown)


mattyx14 (approved):

Sargitarium:
Thx for helping BUT, have this error when click change description on guild menu:

Call to undefined function bcsub()
Backtrace:
#0 /var/www/html/system/pages/guilds.php(30): require()
#1 /var/www/html/index.php(360): require('/var/www/html/s...')
#2 {main}

majestyotbr:
?

slawkens:
@Sargitarium it actually doesn't have anything to do with this PR.

You are missing the bcmath extension. 

How to install? see here
https://stackoverflow.com/questions/51094183/fatal-error-uncaught-error-call-to-undefined-function-bcadd

slawkens:
This can be merged.

majestyotbr:
Thanks.

PR #940, opened 2020-03-24
Fix market imbued item

### LINKED ISSUE #797: Market Imbuiment Item
Itens que já foram imbuidos anteriormente, mesmo com o imbuement removido, 
não são negociáveis no market.

FakeShinoda:
That's it.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -1956,8 +1956,6 @@ bool Item::hasMarketAttributes() const
 			if (duration != getDefaultDuration()) {
 				return false;
 			}
-		} else {
-			return false;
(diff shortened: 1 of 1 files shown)


gerotib:
Tested, working. I mean its possible to create offer of item with cleared imbuement, items with active imbuements are not, as expected I think.

FakeShinoda:
> 
> 
> Tested, working. I mean its possible to create offer of item with cleared imbuement, items with active imbuements are not, as expected I think.

Thanks for testing <3
I'll close this PR to make another one with some summons corrections

PR #947, opened 2020-03-27
Game Store - Fix buying item in Shopping Bags

dudztroyer:
This changes fix when buying from store where items with charges came out with shopping bags or it appeared that you can't hold them because of no slots left.

If you have an item that's getting splitted each 100 charges, check the function `isKegExerciseItem`. You might need to add the item ID to this check or change the type of the offer to `OFFER_TYPE_STACKABLE`.

Also, added magic gold converter to the item list which can't be moved out from store inbox.


majestyotbr:
Thanks for your contribution!

PR #948, opened 2020-03-27
Fix some cooldowns being overlapped

dudztroyer:
Update game.cpp to match createCondition subid to hasCondition subid, having different exhaust for different actions working. 

By the repo before PR, if you look to an item and then try to drag them it will give you exhausted. Now, the exausthed is working as expected (independent from each other).

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -2678,7 +2678,7 @@ void Game::playerRequestTrade(uint32_t playerId, const Position& pos, uint8_t st
 		return;
 	}
 
-	if (player->hasCondition(CONDITION_EXHAUST, 1)) {
+	if (player->hasCondition(CONDITION_EXHAUST, 8)) {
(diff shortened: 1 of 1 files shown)


gerotib:
Theres also "exhaust" problem with overlapping potion usage and other actions, but thats caused by [useItemEx](https://github.com/opentibiabr/otservbr-global/blob/622922157dc854dc7481cd1fb0334b1ca8c22e44/src/actions.cpp#L468) used in both. For example you use potion, then immediately try to open corpse, and youll have to wait ~1 sec. On rl theres no delay. That probably should be reported as another issue tho.

dudztroyer:
Sure, it should be another PR... There should be different cooldowns for using items inside player inventory and to outside objects. But for the issue, this fix some exhausted bugs in the current repo...

majestyotbr:
Thanks for the fix!

PR #956, opened 2020-03-29
Quicklooting fixes and updates

dudztroyer:
This fixes some things like adding the main bp as fallback or searching for other able bps.

TODO add spam protection
TODO make player move to the loot before quicklooting, else it will grab from distance the loot
TODO add sendContainer when user clicks on the backpack inside manage loot container
TODO save quick loot list as a string in storage


chriis7:
is it possible to use simultaneously with autoloot?

Issue #724, opened 2019-11-03
Bug related to house. Game crash!


cristianso:
Since I was trying to find out new things, I will edit here everything that I found out.
There is literally every step I did.

**Reproductions step:**
**1-** Download the "branch: develop" of this base;
**2-** Compile on Ubuntu 14.04;
**3-** Download myaac latest version;
**4-** Use schema.sql that is inside this base;
**5-** Set config.lua "houseRentPeriod="weekly" or "monthly" (did not test "daily" or "annual");
**6-** Use the talkaction !buyhouse to buy the house (I've tested tons of talkactions that buy houses, all of them I have the same result);
**7-** Use global map that is inside of this base;
**8-** Start your server;
**9-** Get enough money in your balance and buy a house;
**10-** After you buy the house, restart your server;
**11-** Next time you try to login, your client is going to debug and you have to delete your char from the game. (Sometimes you can login to the game, but you must try tons of times);

**Considerations:** Doesn't matter where your char is or what house you bought, the result is the same. You can buy the house and teleport to anywhere in the map, your game is going to bug when you restart the game.

**Client**: 12
**OS**: Ubuntu 14.04

cristianso:
Seems like a random thing. I tried to redo these steps and this time did not work. There is something more that is triggleling this bug, but I don't know yet.

dudantas:
You may have made a change to your base that is causing this ...
Try to follow the issue recommendations (when you will open it)
Download the latest CLEAN base without any changes. Compile it, use the CLEAN database and see if the debug will occur. Most of these random debugs are not due to any base issues, but to some script you add.

christiangamaesilva:
Bro..... Did you even TRY?! I just compiled for the third time a full clean base, didn't edit anything at all.
IT WILL CRASH! 
Please, before saying that, try before, okay?

christiangamaesilva:
Just change houseRentPeriod from  = "never" to "weekly", or "monthly" and then buy a house ingame, restart the server and try to login again. Your game is going to CRASH! You added the label unconfirmed and didn't even try doing the steps I said....

dudantas:
> Just change houseRentPeriod from = "never" to "weekly", or "monthly" and then buy a house ingame, restart the server and try to login again. Your game is going to CRASH! You added the label unconfirmed and didn't even try doing the steps I said....

You who have to give information. I don't read minds.
When creating the issue has a lot of information, you should tell how you reproduced the bug and not just say you found it.

[...]

PR #958, opened 2020-03-30
Changes in the source and datapack.

dudantas:
Synchronizations with the TFS source/pack and others changes:

* Fixed:
- [x] Bug in droploot
- [x] Bug in monsters for revscryptsys
- [x] Issue in the houses system #724
- [x] Bug at the blessings system (sometimes dropped an item from the set or lost several levels, even with bless)
- [x] Fix for: https://github.com/opentibiabr/otservbr-global/issues/963

* Changes:
- [x] Experience stages moved to Lua (credits for @ranisalt), new file stages in folder "data/stages.lua"
Note: config.lua "rateExp" doesn't change the exp anymore, it's just for fallback
Change the rate / stage experience in the "data/stages.lua" file
- [x] Moved encryption to XTEA
- [x] House functions moved back to housetile.cpp
- [x] Change the speed of all monsters in the XML script. (The speed always needs to be multiplied by the monster real speed.)

* Removed:
- [x] Exhaust for look
- [x] Exhaust for move
- [x] Exhaust for trade
- [x] Exhaust for change outfit
- [x] Exhaust for speak
- [x] Spdlog lib
- [x] Multiplication of 2.5 * for monsters speed in the source

* Added:
- [x] Bless talkaction
- [x] Sigbreak handler for windows
- [x] Revscriptys check register interface

* Added missing functions:
- [x] hasFlag (group)
- [x] getMana (player)
- [x] addMana (player)
- [x] getMaxMana (player)
- [x] hasCondition (creature)
- [x] getMonsterTypes (game)
- [x] addLoot (monsterType)
- [x] registerEvent (monsterType)

Some modifications were just to move functions to other locations, staying as in TFS and making it easier in case someone needs to make a diff between the source OTBR and TFS.

Credits to: 
The Forgotten Server
@dudantas 
@gpedro (for the fixes on the monsters XML speed.)


src/game.cpp
```diff
@@ -5229,69 +5168,6 @@ void Game::loadPlayersRecord()
...
-				stages[lastStageLevel] = multiplier;
-			} else {
-				for (uint32_t i = minLevel; i <= maxLevel; ++i) {
-					stages[i] = multiplier;
-				}
-			}
-		}
-	}
-	return true;
-}
```
Leo32onGIT, line 5293: Are you removing stages.xml?
Users existing code that is referencing stages.xml, will either be stale or broken after this.

i.e. ACC Makers do this

dudantas: Ok. Another script will have to be made for the sites, it is not a difficult world to do. I'm moving the stages to Lua, as you can see here: https://github.com/opentibiabr/otservbr-global/blob/ae3946a94d280c6f9bc0e34be9600463b5cf0ddd/data/events/scripts/player.lua#L24

I did the tests and it seemed ok, better refactored and easier to change.

data/creaturescripts/scripts/others/droploot.lua
```diff
@@ -1,10 +1,10 @@
 function onDeath(player, corpse, killer, mostDamage, unjustified, mostDamage_unjustified)
-	if getPlayerFlagValue(player, PlayerFlag_NotGenerateLoot) then --or player:getVocation():getId() == VOCATION_NONE then
+	if player:hasFlag(PlayerFlag_NotGenerateLoot) or player:getVocation():getId() == VOCATION_NONE then
```
gpedro: players without vocation cannot drop loot?

data/events/scripts/player.lua
```diff
@@ -18,6 +18,32 @@ local STORAGE_CAPACITY_IMBUEMENT = 42154
 -- Players cannot throw items on teleports if set to true
 local blockTeleportTrashing = true
 
+-- Minlevel and multiplier are MANDATORY
+-- Maxlevel is OPTIONAL, but is considered infinite by default
+-- Create a stage with minlevel 1 and no maxlevel to disable stages
+	experienceStages = {
```
gpedro: we cannot move this config to another file?

src/configmanager.cpp
```diff
@@ -106,7 +104,7 @@ bool ConfigManager::load()
 
 	luaL_openlibs(L);
 
-	if (luaL_dofile(L, configFileLua.c_str())) {
```
gpedro: i disagree with that. once we can use a custom config file using argument `-c xablau.lua`

https://github.com/opentibiabr/otservbr-global/blob/93e13351ee12661c84693f208723c9abcd44dff4/src/otserv.cpp#L139-L144

src/connection.cpp
```diff
@@ -265,23 +263,23 @@ void Connection::parsePacket(const boost::system::error_code& error)
 				return;
 			}
 		} else {
-			msg.skipBytes(1); // Skip protocol ID
+			msg.skipBytes(1);    // Skip protocol ID
```
gpedro: trailing spaces

src/connection.cpp
```diff
@@ -265,23 +263,23 @@ void Connection::parsePacket(const boost::system::error_code& error)
...
 		}
 
 		protocol->onRecvFirstMessage(msg);
 	} else {
-		protocol->onRecvMessage(msg); // Send the packet to the current protocol
+		protocol->onRecvMessage(msg);    // Send the packet to the current protocol
```
gpedro: trailing spaces

src/const.h
```diff
@@ -256,8 +253,6 @@ enum MessageClasses : uint8_t {
 	MESSAGE_PARTY = 35, /*White message in channel (+ channelId)*/
 	MESSAGE_EVENT_ORANGE = 36, /*Orange message in the console*/
 	MESSAGE_STATUS_CONSOLE_ORANGE = 37,  /*Orange message in the console*/
-
-	MESSAGE_STATUS_CONSOLE_BLUE = MESSAGE_EVENT_ADVANCE,
```
gpedro, line 260: have you checked all references?

src/protocolold.cpp
```diff
@@ -1,8 +1,6 @@
 /**
- * @file protocolold.cpp
```
gpedro, line 2: protocolold is unused, maybe we can remove this protocol.

src/protocollogin.h
```diff
@@ -1,8 +1,6 @@
 /**
- * @file protocollogin.h
```
gpedro, line 2: protocollogin is unused too

src/game.cpp
```diff
@@ -4990,7 +4930,6 @@ void Game::shutdown()
 {
 	std::cout << "Shutting down..." << std::flush;
 
-	saveGameState();
```
gpedro, line 4993: we need to check if when press ctrl-c to finish the process, he still saving the server

data/lib/compat/compat.lua
```diff
@@ -319,36 +319,36 @@ addOutfitCondition = Condition.setOutfit
...
-function getCreatureOutfit(cid) local c = Creature(cid) return c ~= nil and c:getOutfit() or false end
-function getCreatureSpeed(cid) local c = Creature(cid) return c ~= nil and c:getSpeed() or false end
-function getCreatureBaseSpeed(cid) local c = Creature(cid) return c ~= nil and c:getBaseSpeed() or false end
+function isCreature(cid) return Creature(cid) end
+function isPlayer(cid) return Player(cid) end
+function isMonster(cid) return Monster(cid) end
+function isSummon(cid) return Creature(cid):getMaster() end
+function isNpc(cid) return Npc(cid) end
+function isItem(uid) return Item(uid) end
+function isContainer(uid) return Container(uid) end
```
gpedro: and you double check this method? Methods with prefix `is` should returns boolean and `Creature(cid)` returns an obj

gpedro (changes requested):

data/events/scripts/player.lua
dudantas: Yes we can.
I initially thought about moving to config.lua, but I ran into problems reading login.php and installing the websites.
Any suggestion?

src/game.cpp
dudantas: I'll check.

src/protocolold.cpp
dudantas: Ok

src/const.h
dudantas: @gpedro yes, it's deprecated.

src/game.cpp
dudantas: @gpedro checked, working.

src/configmanager.cpp
dudantas: Reverted

src/protocollogin.h
dudantas: Removed

data/lib/compat/compat.lua
dudantas: Reverted to original changes.

data/events/scripts/player.lua
dudantas: Moved to data/stages.lua and created /reload stages

mattyx14 (approved):
marge to  lasted changed are good very nice work

dudantas:
> marge to lasted changed are good very nice work

Thanks, we still need to correct the speed of the monsters, but soon it will be ready and I will go up so we can merge.

gpedro (approved):
LGTM;

dudantas:
Speed in the XML of the monsters fixed.

worthdavi:
Where is the credits for skillstages creator? :P @LukSrT

slawkens:
To be honest, I don't like that stages are saved in Lua instead of XML.

This will make it harder to parse it by AAC makers.

Why no keep the format of the original stages.xml in TFS as a standard?

PR #959, opened 2020-03-30
Bestiary update

dudztroyer:
Missing charm related functions. Bestiary should be working, tracking kills, announcing when new entry gets unlocked... Giving out charm points as well.

Ive added two access talkactions that helps managing the charm points so it can be easy to debug the following updates.


data/scripts/talkactions/god/charms.lua
```diff
@@ -0,0 +1,56 @@
...
+	if param == "" then
+		param = player:getName()
+	end
+	local target = Player(param)
+	if not target then
+		player:sendCancelMessage("A player with that name is not online.")
+		return false
+	end
+
+	player:sendCancelMessage("Resetado as charms do personagem '" .. target:getName() .. "'.")
```
gpedro, line 50: english pls.

PR #960, opened 2020-03-31
feat: quickloot

### LINKED ISSUE #802: Quick looting
Hi

Is the quick looting system supposed to be working?

I can't get it working, compiled the latest release and using tibia 12 client.

Thanks in advance

gpedro:
- [x] loot corpse [4]
- [x] loot item (on corpse) [4]
- [x] select/remove backpack / gold pouch
- [x] whitelist and blacklist loot mode 
- [x] move items on nested backpack with same id
- [x] main container as fallback
- [x] use function only nearby corpse [1]
- [x] open loot container (needs exports player->addContainer to lua bridge)
- [x] loot analyser (needs exports player->updateLootTracker to lua bridge)
- [x] drop analyser (needs exports player->updateKillTracker to lua bridge)
- [x] container icons
- [x] handle move item to remove category out of inventory [2]
- [ ] loot messages [3]
- [ ] add main bp as unassigned

---
1 - missing autowalk if corpse is far than 1sqm
2 - when you drop bp outside inventory, the attribute is removed but for some reason is not removing from backpack selection window
3 - damn, this has many and many types of messages. idk where to start
4 - if you have 10 gold in backpack and dropped 20gp, if your BP is with slots full, he moves to next container (if have). currently is not stacking when moves


edmararulez:
it is possible to get loot over long distances

VitorNM:
The loot/hunt analysers and loot tracker are not tracking the quick-looted items.

The issue with quick looting by distance was resolved, but when you right-click in a distance corpse, the character doesnt move to the corpse. You need to manually walk to close to the corpse to loot it.

gpedro:
@VitorNM loot analyser and drop tracker fixed

Flyckks:
nossa, parece muito com o sistema que a OTG vai ter daqui uns dias 
¯\_(ツ)_/¯

genrider:
Não estou neste tópico e não sei do que se trata a conversa.

majestyotbr:
Comentários desnecessários removidos, favor ler:
https://github.com/opentibiabr/otservbr-global/blob/develop/CODE_OF_CONDUCT.md

luucassoares:
Hello. I have tested it and so far quickloot is working pretty good. The only thing I noticed it is that when you have a groupable item like small emerald and its container gets full, the next small emeralds I collect will be placed in the next bp instead the first even the group isnt with 100 items.

But it is working really really good

dudztroyer:
@luucassoares i've noticed this aswell. But the way we implemented was to be easy to go to new backpacks when needed, just searching for empty slots.

If we check every item in each backpack set up to see if their stack is lower than 100, it would get a massive performance loss, since we would have to split up the stack between multiple backpacks...

Maybe someday tho

matheuspergher:
Sorry guys, is Autloot working? How is the command?

Thx

PR #961, opened 2020-03-31
[feat] colored loot message

vagnerkuntz:


gpedro:
i know the implementation is correct, but which version does it work from?

vagnerkuntz:
@gpedro Tested only in version 12

gpedro:
@vagnerkuntz `12.??` which minor version?

vagnerkuntz:
To be more specific, it is the same as there are repository tools: 
https://github.com/opentibiabr/tools/blob/master/Tibia%20Client%2012.rar

it is possible that it works with some previous versions / minor versions.
I can't confirm it, I'm still at work.

gpedro:
image
image

aurelion5670:
Version 12.007695, it isn't working either here, also it needs to have a check on the clientversion to avoid the issue that gpedro is having now

FakeShinoda:
This feature has been implemented in version "12.00.7695", the problem is the conflict within these 2 contenctDescriptions functions lua/src.

Otherwise if you close this PR, i can make a another with this function working (only for 12+)

PR #964, opened 2020-03-31
Continued work in bestiary and charms

dudztroyer:
Now it blacks out the monster when you dont have it unlocked yet. Charms can be unlocked. There is a charm expansion method aswell, which someone should link to the gamestore (player:setCharmRuneSlotExpansion(true/false)). There is also a method that you can use to get the active charm for the monster(if any): player:getCharmFromTarget(target)

TODO add the methods to trigger the charm runes
TODO add information related to monster locations in bestiary
TODO add information of loot form special event (this should be fixed in creature loots and then parsed to the bestiary)

Special thanks to Rick and lBaah :)


dudztroyer:
My PR description is in the commits here...

PR #966, opened 2020-03-31
Improved the handling of adding experience

dudztroyer:
This way it wont loop over all destination levels. In the old way the server would freeze until it calculates the final exp. Now, it only makes one calculation and gives the remaining xp.

### DIFF
--- a/data/scripts/talkactions/god/add skill.lua
+++ b/data/scripts/talkactions/god/add skill.lua
@@ -55,13 +55,20 @@ function talk.onSay(player, words, param)
 	end
 
 	local ch = split[2]:sub(1, 1)
-	for i = 1, count do
-		if ch == "l" or ch == "e" then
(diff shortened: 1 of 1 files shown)


data/scripts/talkactions/god/add skill.lua
```diff
@@ -55,13 +55,20 @@ function talk.onSay(player, words, param)
 	end
 
 	local ch = split[2]:sub(1, 1)
-	for i = 1, count do
-		if ch == "l" or ch == "e" then
-			target:addExperience(getExpForLevel(target:getLevel() + 1) - target:getExperience(), false)
-		elseif ch == "m" then
+	print(ch)
+	print(count)
```
gpedro: whoops

majestyotbr:
Tested and working. Thanks for your contribution.

PR #976, opened 2020-04-02
[WIP] feat: supply stash 

dudztroyer:
PLEASE FOR WHOEVER IS GOING TO LOOK AT THIS: HELP US FIND HOW TO INSERT ITEMS INTO THE STASH BY ITEMS CONTEXT MENU. IT HAS SOMETHING TO DO WITH THE FLAGS SENT VIA ADDITEM.

Also, if anything just give me a call.


marcosvf132:
All items are crashing client except for sniper arrow, white mushroom and teleport crystal. Probably the problem is not on the script, but on the client

PR #1008, opened 2020-04-10
Removing unused columns.

omarcopires:


slawkens:
LGTM.

Just not sure about player_charms table. you sure it's not used anywhere?

dudztroyer:
Not used by the bestiary... I built it up using only storages... Didn't check if other stuff is using it tho

omarcopires:
> LGTM.
> 
> Just not sure about player_charms table. you sure it's not used anywhere?

All columns and tables have been checked, there is no use for any of them.

PR #1027, opened 2020-04-15
Improves checkers

Costallat:
Add checker for c++

Improves lua checker to real check files

Clean builder files

Improves checker to only check chaged or added files.


majestyotbr:
Thanks.

PR #1029, opened 2020-04-15
Market Accepting Cleared Imbuiment Items

omarcopires:


src/item.cpp
```diff
@@ -1953,16 +1953,19 @@ bool Item::hasMarketAttributes() const
...
-		}
-	}
-
-	if (items[id].imbuingSlots > 0) {
-		for (uint8_t slot = 0; slot < items[id].imbuingSlots; slot++) {
-			Item* item = const_cast<Item*>(this);
-			uint32_t info = item->getImbuement(slot);
-			if (info) {
-				return false;
+		} else if (attr.type == ITEM_ATTRIBUTE_CUSTOM) {
```
gpedro: why need that if imbuements have our own imbuements attribute?
https://github.com/opentibiabr/otservbr-global/blob/37f01d936575d048ad0b63cef7b381279a69cdb2/src/enums.h#L93

PR #1042, opened 2020-04-17
Removes spdlog in place of a builtin log lib

Costallat:


src/loguru.cpp
```diff
@@ -0,0 +1,1873 @@
...
+	{
+		auto name = s_verbosity_to_name_callback
+				? (*s_verbosity_to_name_callback)(verbosity)
+				: nullptr;
+
+		// Use standard replacements if callback fails:
+		if (!name)
+		{
+			if (verbosity <= Verbosity_FATAL) {
+				name = "FATL";
```
gpedro, line 857: FATL or FATAL?

Costallat: FATL

PR #1053, opened 2020-04-18
fixing hidden creature health

### LINKED ISSUE #86: monsters with hidehealth is not hidding health
hidehealth tag is not workingn properly, it is showing creature's hp bar in black to all players (usually only gms+ can see that)

It is supposed to hide both the creature hp bar and name, but instead it should everything and the hp bar is black

currently:

image

how it's supposed to show:

image

### LINKED ISSUE #962: monsters with hidehealth is not hidding health
hidehealth tag is not workingn properly, it is showing creature's hp bar in black to all players (usually only gms+ can see that)

It is supposed to hide both the creature hp bar and name, but instead it should everything and the hp bar is black

currently:

image

how it's supposed to show:

image

jlcvp:
fixes #962


jlcvp:
@gpedro pode testar esse commit?

src/enums.h
```diff
@@ -134,6 +134,7 @@ enum CreatureType_t : uint8_t {
 	CREATURETYPE_SUMMONPLAYER = 3,
 	CREATURETYPE_SUMMON_OWN = 3,
 	CREATURETYPE_SUMMON_OTHERS = 4,
+    CREATURETYPE_MONSTER_HIDDEN = 5,
```
gpedro: troca espaço por tab, só pra manter o codestyle neste arquivo

src/luascript.cpp
```diff
@@ -1493,6 +1493,7 @@ void LuaScriptInterface::registerFunctions()
 	registerEnum(CREATURETYPE_SUMMONPLAYER)
 	registerEnum(CREATURETYPE_SUMMON_OWN)
 	registerEnum(CREATURETYPE_SUMMON_OTHERS)
+    registerEnum(CREATURETYPE_MONSTER_HIDDEN)
```
gpedro: troca espaço por tab, só pra manter o codestyle neste arquivo

src/protocolgame.cpp
```diff
@@ -3396,7 +3396,9 @@ void ProtocolGame::sendModalWindow(const ModalWindow& modalWindow)
 void ProtocolGame::AddCreature(NetworkMessage& msg, const Creature* creature, bool known, uint32_t remove)
 {
 	CreatureType_t creatureType = creature->getType();
-
+    if (player->getProtocolVersion() >= 1120 && creatureType == CREATURETYPE_MONSTER && creature->isHealthHidden()) {
```
gpedro: isso vai cagar os emblemas lá em baixo, porque ele não vai reconhecer como MONSTER e não vai adicionar os emblemas de summons.

talvez uma flag pra lançar ou algo do tipo...

gpedro (changes requested):
funcionou, mas precisa ver o lance dos emblemas para não dar problemas.

src/protocolgame.cpp
jlcvp: Monstros hidden health não são summonáveis

src/luascript.cpp
jlcvp: done

src/protocolgame.cpp
jlcvp: refiz a lógica pra não alterar o type (apesar do if do summon alterar), de qualquer forma se um monstro hidden for sumonável algum dia, ele não vai ter badge né? senão deixa de ser hidden

jlcvp:
corrigi os pontos, no client 12.30 ficou aparecendo o nome da criatura, coloquei uma verificação pra enviar string vazia

src/protocolgame.cpp
```diff
@@ -3406,7 +3405,11 @@ void ProtocolGame::AddCreature(NetworkMessage& msg, const Creature* creature, bo
 		msg.add<uint16_t>(0x61);
 		msg.add<uint32_t>(remove);
 		msg.add<uint32_t>(creature->getID());
-		msg.addByte(creatureType);
+        if (player->getProtocolVersion() >= 1120 && creatureType == CREATURETYPE_MONSTER && creature->isHealthHidden()) {
+			msg.addByte(CREATURETYPE_MONSTER_HIDDEN);
```
gpedro: tabs pls, vou testar e já te falo se deu certo

gpedro (changes requested):

guiismiti:
Still not working:
1 - setting hidden health with setHiddenHealth:
Players can still see the name and health bar of a god using /hide:
god_hide_health

2 - when killing a monster, the health bar and name are not disappearing immediately:
ezgif com-optimize (1)

Behavior in client 10 (correct behavior 1) - the name and healthbar disappear immediately, before the corpse hits the ground:
ezgif com-optimize (2)

Behavior in real Tibia (correct behavior 2) - the corpse hits the ground immediately:
ezgif com-optimize (3)

guiismiti:
maybe something in sendCreatureHealth in protocolgame.cpp?

jlcvp:
Hi guys, done some more changes to address the hidden gamemaster / god / accessplayer problem. 

@guiismiti I've done an adjust regarding not sending health of hidden health creatures at all, but I think a more definitive solution would be to rewrite the creature dying logic.

The client 10 automatically hides the healthbar if a value of 0 is sent. clients 11/12 don't. 

On the TFS when we kill a creature, first we send health 0 and in the next tick we send the remove creature/body logic. this needs to be changed.

Since it's unrelated to #962 I would recommend opening a different issue regarding this behavior.

gpedro:
@jlcvp https://otland.net/threads/sendcreaturehealth-issue-with-clients-11.270003/

jlcvp:
@gpedro I've sent a more in-depth response in the otland topic. But I'll stick with my opinion that the empty creature health on death is not related with this particular issue. #962 is a complain about the healthbar and name showing for creatures with hiddenHealth flag/status. I recommend we create another issue to investigate the possibilities of redoing the monster death routine

gpedro:
image

Zbizu:
does the monster wait 1s for onThink to drop the corpse?
https://github.com/opentibiabr/otservbr-global/blob/develop/src/creature.h#L72

PR #1055, opened 2020-04-19
Prevent debug on client 11 after first mob kill

Oen44:


data/modules/scripts/bestiary/bestiary.lua
```diff
@@ -554,7 +553,9 @@ function Player.addBestiaryKill(self, monsterID) --MonsterID can be Name
 	local curCount = self:getBestiaryKillCount(monsterID)
 	local monster = Bestiary.Monsters[monsterID]
 	if curCount == 0 then
-		self:sendBestiaryEntryChanged(monsterID)
+		if self:getClient().version == 1200 then
```
gpedro: should be `>= 1200`, uhm?

Oen44: Is there client higher than 12? Qt client is 1200.

gpedro: we don't support +12 yet, but probably this packet is used in versions higher than 12...

Oen44: Updated

gpedro (approved):

PR #1056, opened 2020-04-19
setBankBalance updated to use uint64

Oen44:
Player balance is using unsigned int64 in source and database.

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -9652,13 +9652,7 @@ int LuaScriptInterface::luaPlayerSetBankBalance(lua_State* L)
 		return 1;
 	}
 
-	int64_t balance = getNumber<int64_t>(L, 2);
-	if (balance < 0) {
-		reportErrorFunc("Invalid bank balance value.");
-		lua_pushnil(L);
-		return 1;
-	}
-
+	uint64_t balance = getNumber<uint64_t>(L, 2);
 	player->setBankBalance(balance);
 	pushBoolean(L, true);
 	return 1;
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -9652,7 +9652,14 @@ int LuaScriptInterface::luaPlayerSetBankBalance(lua_State* L)
 		return 1;
 	}
 
-	uint64_t balance = getNumber<uint64_t>(L, 2);
+	std::string value = getString(L, 2);
+	int64_t signedBalance = std::strtoll(value.c_str(), NULL, 10);
+	uint64_t balance = 0;
+
+	if (signedBalance > 0) {
+		balance = std::strtoull(value.c_str(), NULL, 10);
+	}
+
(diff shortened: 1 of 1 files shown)


src/luascript.cpp
```diff
@@ -9652,13 +9652,7 @@ int LuaScriptInterface::luaPlayerSetBankBalance(lua_State* L)
 		return 1;
 	}
 
-	int64_t balance = getNumber<int64_t>(L, 2);
-	if (balance < 0) {
```
Costallat, line 9656: can you keep that part? or comment why you removed it?

Tks

Oen44: Unsigned can't be less than 0 so it's unnecessary. If I first check with int64 then loss of data.

gpedro: so, if I put a negative value in the state, theoretically the  value should be zero?

Oen44: Not just theoretically, it will be 0. From `player.h` and from `players` table in database.
```cpp
uint64_t bankBalance = 0;
```
```sql
`balance` bigint(20) UNSIGNED NOT NULL DEFAULT '0',
```
There is a huge difference between signed and unsigned int.
I'm making this change because I was informed by a server owner that `Invalid bank balance value.` is thrown very often.
If negative value is passed to `setBankBalance` then something went wrong in calculations.

gpedro:
image

commands used:
- /b 0
- /b 123
- /b -123

script:
```lua
-- data\scripts\talkactions\god\broadcast.lua
local talk = TalkAction("/b")

function talk.onSay(player, words, param)
	if not player:getGroup():getAccess() or player:getAccountType() < ACCOUNT_TYPE_GOD then
		return true
	end
	
	if param == "" then
		player:sendCancelMessage("Command param required.")
		return false
	end

	beforeParam = "before - (bank balance: " .. player:getBankBalance() .. ")"
	player:setBankBalance(param)
	afterParam = "after - (bank balance: " .. player:getBankBalance() .. ")"

	print("> " .. player:getName() .. " broadcasted: \"" .. param .. "\".")
	for _, targetPlayer in ipairs(Game.getPlayers()) do
		targetPlayer:sendPrivateMessage(player, beforeParam, TALKTYPE_BROADCAST)
		targetPlayer:sendPrivateMessage(player, afterParam, TALKTYPE_BROADCAST)
	end
	return false
end

talk:separator(" ")
talk:register()

```

src/luascript.cpp
gpedro: in the practice, I got different results

Oen44:
This is pointless, I'm closing this. I made this based on how often `Invalid bank balance value`  error is thrown. But it's not because the numbers are too small to handle (int64 is pretty damn big), it's that there is something wrong with math that calculates value passed to `setBankBalance`.

PR #1060, opened 2020-04-20
Fix getMinMaxValues attackSkill with distance weapons

Oen44:
When player is using distance weapon that needs ammunition and have ammo item in ammunition slot, attackSkill value is 0, this is because ammunition item is checked first and these items don't have any skill types.

### DIFF
--- a/data/spells/scripts/attack/ethereal spear.lua
+++ b/data/spells/scripts/attack/ethereal spear.lua
@@ -4,10 +4,9 @@ combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_HITAREA)
 combat:setParameter(COMBAT_PARAM_DISTANCEEFFECT, CONST_ANI_ETHEREALSPEAR)
 combat:setParameter(COMBAT_PARAM_BLOCKARMOR, 1)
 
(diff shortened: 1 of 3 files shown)


Leo32onGIT:
Whats the symptom that this fixes?
0 exori con damage in specific scenarios?

Oen44:
> 
> 
> Whats the symptom that this fixes?
> 0 exori con damage in specific scenarios?

Numbers too low. `CALLBACK_PARAM_SKILLVALUE` callback arguments are `player, skill, attack, factor`, now it's using only first 3, however they are misused, `player, attack, factor`. So `attack` is actually `skill`, which means it's 0 if using distance weapon with ammo and `factor` is `attack`.

jlcvp:
LGTM

jlcvp (approved):

lgrossi (commented):
LGTM, however needs to check why builds are failing.

dudantas:
> LGTM, however needs to check why builds are failing.

They are just warnings, indentation, line break, etc.

dudantas (approved):

lgrossi (approved):

dudantas:
Thanks for your contribution!

PR #1066, opened 2020-04-20
Fixed doSetCreatureOutfit to use outfit parameter

Oen44:


gpedro:
what's the difference between lookType and lookTypeEx?

Oen44:
> 
> 
> what's the difference between lookType and lookTypeEx?

lookTypeEx is used if you want to use Item as outfit.
`doSetCreatureOutfit` has parameter `outfit` which wasn't used in code and there is `itemType` which is always `nil`.

gpedro (approved):

Issue #1070, opened 2020-04-22
Crash on market packet attack


jlcvp:
## CAUTION: OUR SERVER IS UPDATED DAILY. MAKE SURE YOU ARE REPORTING SOMETHING ON THE LATEST VERSION OF THE SERVER.

**Describe the bug**
There's an edge case where the coin balance update crashes the server with a null pointer exception

**To Reproduce**
1. Create one or more market offers
2. Use a bot to flood packets of cancellation/buy of these market offers
3. Get kicked from exceeding the packets per second limit
4. crash on ProtocolGame::UpdateCoinBalance

**Expected behavior**
Coins get updated normally

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: MacOS Catalina

PR #1071, opened 2020-04-22
Fixes opentibiabr/otservbr-global#1070

jlcvp:


mattyx14 (approved):

gpedro:
if it doesn't have a `client` or `client->player`, we can disconnect the client to avoid receive other packets, right? what are you think about that?

jlcvp:
@gpedro I think the disconnection is handled elsewhere, this is an edge case of concurrency where the task is sent to the taskpool while the client is yet connected, and in the meantime between the task come to execution the player was kicked (and thus, player pointer is deallocated) and the client reference still exists and is in the proccess of being deallocated.

gpedro (approved):

Leo32onGIT (approved):

PR #1081, opened 2020-04-27
Create recompile.sh

andersonfaaria:
creating sugarcoat for recompiling.
For now we are assuming the default folder structure but later we can turn this into a variable that checks the folder from where the file is (for windows)


Costallat:
Merged on PR #1207

PR #1092, opened 2020-04-28
fix: creature update walk

gpedro:
this change makes onCreatureMove run the task to update follow path on every creature movement

https://imgur.com/a/kP3LQEV

note: probably will increase the CPU usage
credits to @marksamman [(reference)](https://otland.net/threads/weird-monster-behaviour-in-tfs-1-1.228021/post-2200365)


lBaah:
Please describe better what are you doing there, consagrado

PR #1097, opened 2020-04-29
refactoring/fixing reward_common.lua's logic

### LINKED ISSUE #1098: Some bugs and missing features in reward_common.lua
@MajestyOTBR abri o PR https://github.com/opentibiabr/otservbr-global/pull/1097, lá explico as alterações

alisonjf:
- Stackable items with count > 1 now shows following message: _"You have found X PLURALNAME"._
- Checking if players has backpack to receive the reward
- Checking if player's backpacks have empty room to receive the item
- Message is now well formatted
- Item name was not beeing displayed before, the function **"getItemName"** was being called with wrong variable as parameter
- Method extraction and reuse
- **ItemType(itemid):getCharges()** was never false, because when item doesn't have charges, it returns 0, not false.
- Single items message now displays correct article, example: _an orange_ or _a dwarven shield_


andersonfaaria:
@alisonjf this code definetely works but you could achieve half of those changes by changing player:addItem to player:addItemEx and checking the return code

alisonjf:
> @alisonjf this code definetely works but you could achieve half of those changes by changing player:addItem to player:addItemEx and checking the return code

I know bro, I did some tests and when no room it returns 2 and no cap it returns 21, but I did it with addItem cuz I don't particularly like the way the function addItemEx works, but it sure would be cleaner, but in my opinion, not more readable

data/scripts/actions/quests/system/reward_common.lua
```diff
@@ -3,33 +3,51 @@
...
+						return true
 					end
+					player:addItem(itemid, count)
+					player:sendTextMessage(MESSAGE_EVENT_ADVANCE, message .. ".")
+					player:setStorageValue(setting.storage, 1)
 				else
-					player:sendTextMessage(MESSAGE_EVENT_ADVANCE, "You have found a " .. getItemName(setting.itemReward) .. ". Weighing " .. getItemWeight(setting.itemReward) .. " oz it is too heavy.")
+					local itemArticle = itemDescriptions.article
+					local message = "You have found " .. itemArticle .. " " .. itemName
+					if not (checkWeightAndBackpackRoom(player, getItemWeight(itemid), message)) then
```
gpedro: unnecessary parenthesis

data/scripts/actions/quests/system/reward_common.lua
```diff
@@ -3,33 +3,51 @@
...
+				local count = setting.itemReward[i][2]
+				local itemDescriptions = getItemDescriptions(itemid)
+				local itemName = itemDescriptions.name
+
+				if count > 1 and (ItemType(itemid):isStackable() or ItemType(itemid):getCharges() > 0) then
+					if (itemDescriptions.plural) then
+						itemName = itemDescriptions.plural
+					end
+					local message = "You have found " .. count .. " " .. itemName
+					if not (checkWeightAndBackpackRoom(player, getItemWeight(itemid)*count, message)) then
```
gpedro: unnecessary parenthesis

data/scripts/actions/quests/system/reward_common.lua
gpedro: here we have a duplicated/similiar piece of code, we can merge it?

data/scripts/actions/quests/system/reward_common.lua
```diff
@@ -3,33 +3,51 @@
 -- Remember to add the uniqueid at the end of the script.
 local commonReward = Action()
 
+local function checkWeightAndBackpackRoom(player, itemWeight, message)
+	local backpack = player:getSlotItem(CONST_SLOT_BACKPACK)
+	if not backpack or backpack:getEmptySlots(true) < 1 then
+		player:sendTextMessage(MESSAGE_EVENT_ADVANCE, message .. ", but you have no room to take it.")
+		return false
+	elseif (player:getFreeCapacity() / 100) < (itemWeight) then
```
gpedro: unnecessary parenthesis at itemWeight

gpedro: unnecessary "elseif" above, cuz if this condition is valid, he will return and will not reach here

gpedro (changes requested):

data/scripts/actions/quests/system/reward_common.lua
alisonjf: > unnecessary "elseif" above, cuz if this condition is valid, he will return and will not reach here

Its a matter of style, the generated byte code is the same, anyway I'll change it

ikaro1997 (approved):

data/scripts/actions/quests/system/reward_common.lua
```diff
@@ -3,38 +3,63 @@
 -- Remember to add the uniqueid at the end of the script.
 local commonReward = Action()
 
+local function checkWeightAndBackpackRoom(player, itemWeight, message)
```
dudantas: It looks great.
But, maybe we can move this function to 051-functions? So it can be called in other scripts. Who and says?

dudantas (approved):

dudantas:
Good work, thanks for you contribution.

dudantas:
Hello, sorry for the delay in replying. The issue was closed for being unconfirmed. I did the tests and couldn't reproduce.
Make sure that:
You are using the updated datapack, source and database.
Just in case, download everything from 0 and test again.

If you are still able to reproduce the problem. Explain exactly how to reproduce, step by step, what was done. Thus, we can investigate, the information you gave is vague and so you can't do much.

Issue #726, opened 2019-11-04
Missions that need to be developed or revised


dudantas:
Quests that need to be developed where NPC's or scripts are present incompletely or nonexistently.
If you know of any more, please leave it in the comments I will add in no issue. According to what we are fixing, we will mark them as resolved.
Thanks to all those who reported bugs in the quests, but for the sake of organization and ease of finding issues, it is best to stay that way.

- [ ] (Revision) - Grimvale Quest

- [ ] (Revision) - The Ancient Tombs Quest

- [ ] (Revision) - The Shattered Isles Quest

- [ ] (Revision) - Waterskin of Mead Quest

- [ ] (Revision) - Power Bolts Quest

- [ ] (Revision) - Blood Herb Quest

- [ ] (Revision) - Rottin Wood and the Married Men Quest

- [ ] (Revision) - Panpipe Quest
The hollow stone should key 4055 after using it.
(Position: 32652, 32107, 7)
Bounty chest is empty. Add the rewards: Quest box containing a Bag with 1 Power Ring, Panpipes and 2 Small Amethysts
(Position: 32644, 32131, 8)

- [ ] (Develop) - Rest in Hallowed Ground Quest
Need the script for the tombs
AID from 4210 to 4224
Positions(33218, 31669, 7) and (33202, 31685, 8)

- [ ] (Develop) - The Secret Library Quest

- [ ] (Develop) - The Dream Courts Quest

- [ ] (Develop) - The Paradox Tower Quest

- [ ] (Develop) - Twenty Miles Beneath the Sea Quest

- [ ] (Develop) - Threatened Dreams Quest

- [ ] (Develop) - Sinatuki Quest

- [ ] (Develop) - Cartography 101 Quest

- [ ] (Develop) - Krailos  Quest

- [ ] (Develop) - The Cursed Crystal Quest

- [ ] (Develop) - Blood Brothers Quest

- [ ] (Develop) - The Hunt for the Sea Serpent Quest

- [ ] (Develop) - Treasure Hunt Quest

- [ ] (Develop) - Machinery of War Quest

- [ ] (Develop) - The Paradox Tower Quest

- [ ] (Develop) - The Isle of Evil Quest

- [ ] (Develop) - To Blind the Enemy Quest

- [ ] (Develop) - The Repenters Quest

- [ ] (Develop) - Top of the City Quest

- [ ] (Develop) - Goblin Merchant Quest

[...]

PR #1101, opened 2020-04-29
Full Meriana Quest

YohanAugusto:
After completing Eleonore Quest (The Shattered Isles Quest), NPC Raymond Striker couldnt start the quest.

I have added Meriana Quest and should now be working fine. I also changed a few doors on Nargor, which was locked doors instead of closed doors.

If everything is ok, should check Meriana Quest revision on #726.


dudantas:
Merged, thanks for your contribution.

PR #1116, opened 2020-05-01
Fix build on GNU/Linux Ubuntu 16.04 with GCC 5.4.0 and ccache enabled. Closes #1115

### LINKED ISSUE #1115: Error during compilation on Ubuntu 16.04
**Describe the bug**
When compiling I get errors due to the inclusion of std algorithm library. When analyzing I discovered that GCC was trying to use an undefined function for IA32 platform while I'm using Ubuntu 16.04 x86_64.

**To Reproduce**
1. Just clone the updated repository and try to build from command line using GCC 5.4.0.

**Expected behavior**
Compilation starts and finish with success.

 - OS: Elementary OS
 - Version: 0.4 Loki (Ubuntu 16.04 based)

**Additional context**
This only happens when using `ccache` with `gcc` and `-march=native` option. Cannot confirm if this also affects `clang` because during compilation it consumes all RAM available forcing excessive swapping and freezing the system.

jsapparts:
Something was causing the -march to be incorrect, causing inclusion of files from different platforms such as IA32.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,11 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   # https://kristerw.blogspot.com/2017/09/useful-gcc-warning-options-not-enabled.html
 
   # cmake -DCMAKE_BUILD_TYPE=Release ..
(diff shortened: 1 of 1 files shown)


Costallat:
It's better to put this on the wiki as it is related to elementary OS, and all the builds for ubuntu are working properly

Costallat:
https://github.com/opentibiabr/otservbr-global/wiki/Build-on-elementaryOS

PR #1123, opened 2020-05-02
Improvements in ubuntu build

omarcopires:
Code taken from: https://github.com/otland/forgottenserver/pull/3000/files


Costallat:
Parece uma cópia&cola daqui https://github.com/otland/forgottenserver/pull/2973/files#diff-a83fb732088b4692366f86f5bbdae225

Nós podemos fazer melhor que isso e já fiz no #1127 

Tks, Steve Job Jr. for politeness

PR #1125, opened 2020-05-02
fix problem when loading npc_spawn_list.lua

YohanAugusto:
**Problem**

There was a problem when loading all NPC on startup.

Take a look at the function createNpc() from data/globalevents/scripts/spawn/npcs_spawn.lua:

`local function createNpc()
	for i = 1, #mapNpcs do
		local npc = mapNpcs[i]
		if npc and npc.name and npc.pos then
			local spawn = Game.createNpc(npc.name, npc.pos)
			if spawn then
				spawn:setMasterPos(npc.pos)
			end
		end
	end
	return true
end`

It starts loading NPC from index 1, but our list starts from index 2.

**Fix:**

Move the last NPC on the list to index 1.


YohanAugusto:
@dudantas will make another PR with more fixes

dudantas:
Thanks, fixed on https://github.com/opentibiabr/otservbr-global/pull/1126

PR #1130, opened 2020-05-02
Fix: bifgoot burden (warzone variable location)

dudantas:
Moved to quest folder on libs and fixed variables names


ikaro1997 (approved):
Ok!

Issue #1133, opened 2020-05-03
Missing parameter on DLOG_F


syd-sev:
Hi, I was having a crash during the startup running on Ubuntu 18.04 on Amazon, found out it was a missing parameter on the following line:

actions.cpp:662	`DLOG_F(INFO, "Event[%s] configured with success!");`

Funny thing, that line was not causing the server to crash when running it locally (VMware) on a Ubuntu 18.04.4 sever, in fact it was printing:

`"Event[null] configured with success!". However, `

My guess is Amazon is shipping a custom libc, or compilers... 

Thanks,

PR #1138, opened 2020-05-04
Adjust actions logs

Costallat:
- Moved logs to debug

- Fixed some logs with wrong number of 'prints'


Costallat:
#1133

PR #1158, opened 2020-05-07
Print duplicate storages

andersonfaaria:
This change has very low impact on startup and allow people to identify duplicate storages as long as all storages used in the game are inside this table.

### DIFF
--- a/data/lib/core/storages.lua
+++ b/data/lib/core/storages.lua
@@ -2187,7 +2187,7 @@ GlobalStorage = {
 
 }
 
--- Função de extração de values
+-- Values extraction function
 local function extractValues(tab)
   local ret = {}
   for _, v in pairs(tab) do
@@ -2204,16 +2204,16 @@ local function extractValues(tab)
 end
 
 local benchmark = os.clock()
-local extraction = extractValues(Storage) -- Chama a função
-table.sort(extraction) -- Ordena a tabela
--- A escolha da ordenação se deve ao fato de que ordenação é bem barato O(log2(n)) e aí podemos simplesmente comparar um a um os elementos achando duplicatas em O(n)
+local extraction = extractValues(Storage) -- Call function
+table.sort(extraction) -- Sort the table
+-- The choice of sorting is due to the fact that sorting is very cheap O (log2 (n)) and then we can simply compare one by one the elements finding duplicates in O(n)
 
--- Percorre a tabela extraída verificando se há duplicatas
+-- Scroll through the extracted table for duplicates
 if #extraction > 1 then
   for i = 1, #extraction - 1 do
     if extraction[i] == extraction[i+1] then
       print("Duplicate storage value found: ".. extraction[i])
     end 
   end
-  print(string.format("Processado em %.4f(s)", os.clock() - benchmark))
+  print(string.format("Processed in %.4f(s)", os.clock() - benchmark))
 end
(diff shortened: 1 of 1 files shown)


data/lib/core/storages.lua
```diff
@@ -2186,3 +2186,34 @@ GlobalStorage = {
...
+-- Função de extração de values
+local function extractValues(tab)
+  local ret = {}
+  for _, v in pairs(tab) do
+    if type(v) == "number" then
+      table.insert(ret, v)
+    else
+      local extraction = extractValues(v)
+      for _, k in pairs(extraction) do
+        table.insert(ret, k)
```
jlcvp: I think it's better to use the table indexed by the value of the storage, that would require only 1 iteration on the array thus, O(n)

data/lib/core/storages.lua
```diff
@@ -2186,3 +2186,34 @@ GlobalStorage = {
...
+-- A escolha da ordenação se deve ao fato de que ordenação é bem barato O(log2(n)) e aí podemos simplesmente comparar um a um os elementos achando duplicatas em O(n)
+
+-- Percorre a tabela extraída verificando se há duplicatas
+if #extraction > 1 then
+  for i = 1, #extraction - 1 do
+    if extraction[i] == extraction[i+1] then
+      print("Duplicate storage value found: ".. extraction[i])
+    end 
+  end
+  print(string.format("Processado em %.4f(s)", os.clock() - benchmark))
```
jlcvp: I would recommend to use all the strings and comments in english

data/lib/core/storages.lua
jlcvp: I also think that as this only runs once, there's no need for so much algorithmic analysis and this approach should fit well for most of the systems

andersonfaaria: I'm not sure if I've understood what you said, if you change the table Storage to be indexed by storage you won't have the possibility to referentiate the value in the scripts anymore.

As for the algorithmic analysis, is only to show that this code is ultra fast and can be left running every start up without any delays. it's actually O(nlog2(n) + 2n)

data/lib/core/storages.lua
```diff
@@ -2186,3 +2186,34 @@ GlobalStorage = {
 	XpDisplayMode = 5634
 
 }
+
+-- Values extraction function
+local function extractValues(tab)
```
lgrossi: As discussed, the loop can be improved
```
local function extractValues(tab, ret)
  if type(tab) == "number" then
    table.insert(ret, tab)
  else
    for _, v in pairs(tab) do
      extractValues(v, ret)
    end
  end
end

...

local extraction = {}
extractValues(Storage, extraction) 
```

edit: testing 100 times, there seems to be a 35% time improvement

```
local benchmark = os.clock()
for i = 0, 100 do
	local extraction = {}
	extractValues(Storage, extraction)
end 
print(string.format("Processed in %.4f(s) (lucas)", os.clock() - benchmark))

local benchmark = os.clock()
for i = 0, 100 do
	local extraction = extractValues2(Storage)
end 
print(string.format("Processed in %.4f(s) (anderson)", os.clock() - benchmark))

...

Processed in 0.046968(s) (lucas)
Processed in 0.072279(s) (anderson)
```

lgrossi (changes requested):
Small suggestion on recursion improvement.
Looks good in general, quite useful.

dudantas (approved):
LGTM

lgrossi (approved):
Lgtm

dudantas:
Merged, thanks!

PR #1191, opened 2020-05-09
Allow attack feature from config.lua

lBaah:
This will allow everyone who wants to allow/disallow the attack delay swing through the config.lua without removing any feature.
This is an alternative to https://github.com/opentibiabr/otservbr-global/pull/1175


gabz182:
@lBaah 
actionTimerInterruptAttack = **true** -- should actions interval interrupt attack?

Se deixar "true" os paladinos não conseguem atacar mesmo estando parado sem fazer nada, isso está correto?

PR #1192, opened 2020-05-09
Allow regeneration on PZ through config

lBaah:
This will allow food and item regeneration through config.lua.

true = enable
false = disable

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -19,6 +19,7 @@ monthKillsToRedSkull = 10
 redSkullDuration = 30
 blackSkullDuration = 45
 orangeSkullDuration = 7
+allowProtectionZoneRegeneration = false -- enable/disable food regeneration in protection zone
 
 -- Connection Config
 -- NOTE: maxPlayers set to 0 means no limit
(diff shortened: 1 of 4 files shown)


jlcvp (commented):
As this is a "global" repository, the objective is to maintain it in par with the global tibia.

This is a custom modification and doesn't appears to be right to put it on the basecode

andersonfaaria:
@jlcvp 
Despite being a global project, to aim the public who also like to customize the project we should enable more of those 'pre-configured' improvements that can be easily toogled on/off in config.lua.

The default value however should also be the one that mimics global as it is a global project.
Also @lBaah I think you forgot to include the configmanager as it is failing in the checks.

lBaah:
Thank you for your review @andersonfaaria! I really appreciate that.
The code I sent is from an old version of OTBR. Maybe configmanager isn't declared in this newer versions. I'll check.
Thank you again!

@jlcvp I do understand your demand! Therefore, I think we should offer some features that can be enabled/disabled so we can see few new servers not only focused in real Tibia. In my opinion, if someone wants to play a real thing, they should play the official game. This is only an emulator that uses the official game engine.
If you have any further opinion on this matter I would be glad to hear you! Thank you and have a great day!

andersonfaaria:
reviewed and it looks neat, someone care to test and let me know if it works as intended?

One test that I would like to check is wheter if I'm standing in pz and eat food with it set as true and then /reload config now with it set as false if it would stop the condition or no. If someone does that please let me know the output

PR #1195, opened 2020-05-10
Feat: OTC and 12.x only compatibility

FakeShinoda:
- Now ahead we will only support Client 12.00x and otclient from: https://github.com/opentibiabr/otclient

WARNING: 

Test it before apply.
Truncate your table "prey_slots" in your Database.

Please report here if your client 12.00 crash.


jlcvp:
I think this is a backfire and will considerably drop the adoption of this pack.

Also I think this needs to be more specific as we have protocol changes between 12.05, 12.15, 12.20, 12.30 and 12.31 versions. There's no unified "protocol 12"

gabz182:
Testando o PR, parece funcionar corretamente, até agora não encontrei erros.
Se alguém mais puder ajudar a testar para confirmar.
@Heitortadeu

htc16:
Ok! Só vou terminar o teste no PR que o Du fe...

lgrossi:
Why exactly are we doing that? Doesnt really seems worth and remove some flexibility, like otc compatibility. Not a good shout imho.

dudantas:
> Why exactly are we doing that? Doesnt really seems worth and remove some flexibility, like otc compatibility. Not a good shout imho.

In our predictions, trying to leave a support between 10x / otc / 12.31 will make them have more than 50 "if version" in the source, which is, frankly, not viable in the long run, with each protocol update we would have to release more "if versions" and this will pollute the source too much.

Our idea is to completely remove support for 10x and keep only 12x and otc (obviously we will have to make some changes so that otc is supported along with the merge of this pull request)

dudztroyer:
As a complement of @dudantas comment, for every new feature or different packet structure applied will have a potential incompatibility or even a crash point for older versions. If we drop the compatibility officially we can go beyond the new features without having to take a step back for checking compatibility for each old version.

src/protocolgame.cpp
```diff
@@ -1498,13 +1469,13 @@ void ProtocolGame::sendBlessStatus()
...
 
-	if (player->getProtocolVersion() >= 1120) {
+	if (player) {
 		if (blessCount >= 5) //Show up the glowing effect in items if have all blesses
 			flag |= 1;
 
 		msg.add<uint16_t>(flag);
 		msg.addByte((blessCount >= 7) ? 3 : ((blessCount >= 5) ? 2 : 1)); // 1 = Disabled | 2 = normal | 3 = green
-	}else if (blessCount >= 5) {
+	} else if (blessCount >= 5) {
```
lgrossi: Else if and else not needed anymore, its covered by the blessCount flag in the first case.

dudantas:
> As a complement of @dudantas comment, for every new feature or different packet structure applied will have a potential incompatibility or even a crash point for older versions. If we drop the compatibility officially we can go beyond the new features without having to take a step back for checking compatibility for each old version.

Exact. Now it will be done like this.

dudantas (approved):
LGTM

htc16:
Testado, tudo certo!

PR #1197, opened 2020-05-10
[WIP] Optional Full Light for GMs

andersonfaaria:
With this change you'll be able to define if GMs will always have full light or no

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -3448,7 +3448,10 @@ void ProtocolGame::AddCreature(NetworkMessage& msg, const Creature* creature, bo
 	}
 
 	LightInfo lightInfo = creature->getCreatureLight();
-	msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM))
+		msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	else
+		msg.addByte(lightInfo.level);
 	msg.addByte(lightInfo.color);
 
 	msg.add<uint16_t>(creature->getStepSpeed() / 2);
@@ -3756,7 +3759,10 @@ void ProtocolGame::AddItem(NetworkMessage& msg, const Item* item)
 void ProtocolGame::AddWorldLight(NetworkMessage& msg, LightInfo lightInfo)
 {
 	msg.addByte(0x82);
-	msg.addByte((player->isAccessPlayer() ? 0xFF : lightInfo.level));
+	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM))
+		msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	else
+		msg.addByte(lightInfo.level);
 	msg.addByte(lightInfo.color);
 }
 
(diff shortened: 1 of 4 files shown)


jlcvp (commented):
LGTM, besides the config name induces that an item must be involved

andersonfaaria:
> LGTM, besides the config name induces that an item must be involved

it's because by default you only have full light if you equip an item that provides light or use a spell that gives you light, I didn't know which name to put but feel free to edit the commit

gerotib:
Tested and it works fine. Before I removed that myself, because this full light annoyed me. Tbh I dont even know whats the point of this being in code when you can disable light effects in client.

dudztroyer:
@gerotib it's good to have this toggle since you might want to test map light effects with the GM

gerotib:
@dudztroyer Yeah, thats what I meant. I think that this should work for gms like for normal players, if someone prefer full light all the time then just enable it in client.

andersonfaaria:
Following @gpedro suggesiton I'll move this to a flag instead, I think it makes a lot more sense than a configuration in config.lua. Anyone have a different opinion?

src/protocolgame.cpp
```diff
@@ -3448,7 +3448,10 @@ void ProtocolGame::AddCreature(NetworkMessage& msg, const Creature* creature, bo
 	}
 
 	LightInfo lightInfo = creature->getCreatureLight();
-	msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM))
+		msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	else
+		msg.addByte(lightInfo.level);
```
lgrossi, line 3454: I personally would remove the ternary here:

```
	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM) && player->isAccessPlayer())
		msg.addByte(0xFF);
	else
		msg.addByte(lightInfo.level);
```

That's quite personal and just a suggestion, feel free to ignore.

src/protocolgame.cpp
```diff
@@ -3756,7 +3759,10 @@ void ProtocolGame::AddItem(NetworkMessage& msg, const Item* item)
 void ProtocolGame::AddWorldLight(NetworkMessage& msg, LightInfo lightInfo)
 {
 	msg.addByte(0x82);
-	msg.addByte((player->isAccessPlayer() ? 0xFF : lightInfo.level));
+	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM))
+		msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	else
+		msg.addByte(lightInfo.level);
```
lgrossi, line 3765: same as above

src/protocolgame.cpp
```diff
@@ -3766,7 +3772,10 @@ void ProtocolGame::AddCreatureLight(NetworkMessage& msg, const Creature* creatur
 
 	msg.addByte(0x8D);
 	msg.add<uint32_t>(creature->getID());
-	msg.addByte((player->isAccessPlayer() ? 0xFF : lightInfo.level));
+	if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM))
+		msg.addByte(player->isAccessPlayer() ? 0xFF : lightInfo.level);
+	else
+		msg.addByte(lightInfo.level);
```
lgrossi, line 3778: same as above

src/protocolgame.cpp
```diff
@@ -3766,7 +3772,10 @@ void ProtocolGame::AddCreatureLight(NetworkMessage& msg, const Creature* creatur
...
 	msg.addByte(lightInfo.color);
 }
```
lgrossi, line 3781: Maybe consider to centralize a function 'addLight'? So that we can have single place for the whole light logic.

```
void ProtocolGame::addLight(const OutputMessagePtr &msg, const LightInfo lightInfo)
{
    if(g_config.getBoolean(ConfigManager::GM_FULL_LIGHT_ON_EQUIP_ITEM) && player->isAccessPlayer())
		msg.addByte(0xFF);
	else
		msg.addByte(lightInfo.level);
}
msg.addByte(lightInfo.color);
```

lgrossi (commented):
General suggestions, I strong recommend to centralize the addLight logic, though.

PR #1201, opened 2020-05-10
Improve onKill callback

andersonfaaria:
This change will add another parameter to 'onKill' callback in creaturescript where it would return true if you (or your summon) were the one to perform the finishing blow in the target.

Very useful to avoid bugs in task systems, specially if you consider making them to count also to party members.

By default onKill runs for both mostdamagekiller and lastHit which means 1 kill can count twice

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -618,7 +618,7 @@ void Creature::onDeath()
 	Creature* lastHitCreature = g_game.getCreatureByID(lastHitCreatureId);
 	Creature* lastHitCreatureMaster;
 	if (lastHitCreature) {
-		lastHitUnjustified = lastHitCreature->onKilledCreature(this);
+		lastHitUnjustified = lastHitCreature->onKilledCreature(this, true);
 		lastHitCreatureMaster = lastHitCreature->getMaster();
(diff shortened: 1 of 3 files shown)


gpedro (approved):

src/creature.cpp
```diff
@@ -618,7 +618,7 @@ void Creature::onDeath()
 	Creature* lastHitCreature = g_game.getCreatureByID(lastHitCreatureId);
 	Creature* lastHitCreatureMaster;
 	if (lastHitCreature) {
-		lastHitUnjustified = lastHitCreature->onKilledCreature(this);
+		lastHitUnjustified = lastHitCreature->onKilledCreature(this, true);
```
lgrossi, line 621: You don't really need the true here, since in the .h declaration onKilledCreature has default true for lastHit. However, I'd keep it since it's easier to understand what is going on reinforcing the true value.

lgrossi (approved):
lgtm

src/creature.cpp
andersonfaaria: it was just for making it very explicit that this are the cases where this is true.

Issue #1091, opened 2020-04-28
Gamemasters have logout/login effects even in ghostmode


andersonfaaria:
![unknown (1)](https://user-images.githubusercontent.com/8560144/80546484-1c36c680-898c-11ea-9d0f-f2fff50ef2c7.png)
![unknown](https://user-images.githubusercontent.com/8560144/80546487-1d67f380-898c-11ea-82bb-ff9917c235be.png)

PR #1202, opened 2020-05-10
fixing race condition in outfit

andersonfaaria:
condition invisible trigger a change in outfit (sparkles). They are linked in a way that you're not supposed to be invisible without the outfit. This PR is to guarantee you won't have any outfit change if you're invisible.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -3899,11 +3899,11 @@ void Game::internalCreatureChangeOutfit(Creature* creature, const Outfit_t& outf
 		return;
 	}
 
-	creature->setCurrentOutfit(outfit);
-
 	if (creature->isInvisible()) {
 		return;
 	}
+	
+	creature->setCurrentOutfit(outfit);
 
 	//send to clients
 	SpectatorHashSet spectators;


gpedro:
resolves #1091?

andersonfaaria:
> resolves #1091?

hi @gpedro, I'm afraid no. To fix 1091 you just need to make a check before sending the effects if the player is inghost mode.

This change is to fix another bug that I never seen anyone talking about, when you mix some changes of outfits (swim, stealth ring and changing outfit) you can achieve some weird results like having the swimming outfit outside the water or having regular outfit in water.

gerotib:
Tested, and before this pr I couldnt reproduce any invisibility + water etc bug, with this pr its just: 
1. equip stealth ring
2. walk into water
3. deequip stealth ring = youre "standing" on water and vice versa

So it actually broke that instead of fixing.

andersonfaaria:
> Tested, and before this pr I couldnt reproduce any invisibility + water etc bug, with this pr its just:
> 
> 1. equip stealth ring
> 2. walk into water
> 3. deequip stealth ring = youre "standing" on water and vice versa
> 
> So it actually broke that instead of fixing.

Thank you so much for testing it, I'll take a deeper look in all the invisibility possibilities (utana vid, stealth ring and changing outfit/mounting) before considering closing this PR. The last thing I want is to introduce a new defect hahaha

andersonfaaria:
After some detailed tests I've found out that this is indeed working as expected (before the PR). I'll close this and cancel this pull request.

PR #1213, opened 2020-05-13
Prepare monsters to work on change target strategy and NPC Kendra

htc16:
Change target strategy changing in next commit values added in files are
based on leaked files and NPC Kendra accesses Nostalgia!


gpedro:
can you add a new line between `targetstrategies` and `flags`? this tags is in same line, that's little complicated to read

jlcvp:
> can you add a new line between `targetstrategies` and `flags`? this tags is in same line, that's little complicated to read

@Heitortadeu I think you know how to automate, but here it go:

using vscode, go to the global find/replace, activate the regex (.*) and paste this on find field:
`(<targetstrategies.*/>)\s+(<flags>)` and on the replace field, paste this: `$1\n$2`

click on replace all and Magic!
image

htc16:
Hoje de noite já faço isso.. Valeu!

andersonfaaria:
tested and working

PR #1217, opened 2020-05-13
Fixing Sub type problems

andersonfaaria:
Before this PR:
- No support for item.minCount
- Any item that has more than 100 of count will be restricted for 100.
- Fluid container will always drop with water.
- Npcs won't buy any items with subtypes different from the specified. If none is specified it will only buy 'empty' and if you're trying to sell a golden mug with water it won't appear in the sell list.

After this PR:
- If you edit source and insert item.minCount the method will handle it. Check #1218 for how to implement it.
- Items don't need to be duplicated inside monsters.xml as now you can specify values bigger than 100
- Fluids container (such as golden mug of Dragon Lord) will always drop 'empty'
- Npcs that do not have specific subtypes to buy will buy ANY subtype

### DIFF
--- a/data/lib/core/container.lua
+++ b/data/lib/core/container.lua
@@ -9,20 +9,26 @@ function Container.createLootItem(self, item)
 
 	local itemCount = 0
 	local randvalue = getLootRandom()
+	local lootBlockType = ItemType(item.itemId)
+	
 	if randvalue < item.chance then
-		if ItemType(item.itemId):isStackable() then
-			itemCount = randvalue % item.maxCount + 1
+		if lootBlockType:isStackable() then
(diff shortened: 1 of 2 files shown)


jlcvp (approved):
lgtm

gerotib:
I tested each change along with #1218 and didnt notice any issues.

andersonfaaria:
> I tested each change along with #1218 and didnt notice any issues.

did it solved the proposed problems?

gerotib:
@andersonfaaria yeah, I tested loot with single and duplicated items, with coutmin/max in many different configurations. Monsters doesnt drop fluidcontainers with water, and npc will buy fluidcontainer with any subtype.

PR #1218, opened 2020-05-13
Support for countMin in loot

andersonfaaria:
This PR allows you to use countMin in the definition of a loot.

Now you can have countmin, countmax, both or neither.

After doing this commit please do also #1217 so you have countMin considered in the dropLoot function.

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -1319,6 +1319,12 @@ bool Monsters::loadLootItem(const pugi::xml_node& node, LootBlock& lootBlock)
 	} else {
 		lootBlock.countmax = 1;
 	}
+	
+	if ((attr = node.attribute("countmin"))) {
+		lootBlock.countmin = std::max<int32_t>(1, pugi::cast<int32_t>(attr.value()));
+	} else {
+		lootBlock.countmin = 1;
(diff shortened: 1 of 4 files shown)


src/monsters.cpp
```diff
@@ -1319,6 +1319,12 @@ bool Monsters::loadLootItem(const pugi::xml_node& node, LootBlock& lootBlock)
 	} else {
 		lootBlock.countmax = 1;
 	}
+	
+	if ((attr = node.attribute("countmin"))) {
+		lootBlock.countmin = std::max<int32_t>(1, pugi::cast<int32_t>(attr.value()));
```
jlcvp, line 1324: lootBlock.countmin isn't declared in monsters.h

jlcvp (changes requested):
countmin is not declared on struct LootBlock on monsters.h

src/monsters.cpp
andersonfaaria: you're right, must have missed this one. I'll add as soon as I leave the work

jlcvp (approved):
LGTM

lgrossi (approved):
lgtm

PR #1221, opened 2020-05-14
Fix wrong fight condition

andersonfaaria:
In some specific situations where a player is attacked he did not get infight condition and therefore can just logout/enter pz.

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -3589,6 +3589,8 @@ void Player::onAttackedCreature(Creature* target)
 				pzLocked = true;
 				sendIcons();
 			}
+			
+			targetPlayer->addInFightTicks();
 
 			if (!Combat::isInPvpZone(this, targetPlayer) && !isInWar(targetPlayer)) {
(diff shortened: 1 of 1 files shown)


gabz182:
Não consegui reproduzir o mal funcionamento na atual par poder testar o PR, tudo parece funcionar como deveria.

gerotib:
Also couldnt reproduce that, so cant really test.

andersonfaaria:
> Also couldnt reproduce that, so cant really test.

I'll review the functions, perhaps we may have the addInfight somewhere in the other functions. At least in my server it was bugged before this PR and if we look at TFS code they also have this line there so it made sense to made adding it.

andersonfaaria:
After some detailed tests I've found out that this is indeed working as expected (before the PR). I'll close this and cancel this pull request.

andersonfaaria:
After intensive testing it seems that this bug is not occuring so I'm proceeding to close this PR.

PR #1222, opened 2020-05-14
Fixing login/logout effects for gamemasters

### LINKED ISSUE #98: Gamemasters have logout/login effects even in ghostmode
unknown (1)
unknown

### LINKED ISSUE #1091: Gamemasters have logout/login effects even in ghostmode

andersonfaaria:
Before this PR:
- If you logout in ghost mode everyone sees the poff effect.
- If you login in ghost mode everyone sees the teleport effect.

After this PR:
- If you logout in ghost mode you don't send poff effect
- If you have access or is from an account with access you don't send teleport effect at all (even outside ghost mode*)

*I needed to do it like this because when you are executing 'sendAddCreature' the isInGhostMode() always return false for some reason.

Fix #1091


andersonfaaria:
Tested and working

PR #1234, opened 2020-05-16
Some very useful functions

andersonfaaria:
- New isWalkable function that replaces several other constructed ones and has all the major checks
- getDirectionTo which returns the direction you need to face to start moving from pos1 to pos2
- pairsByKeys which allows you to iterate through tables using keys returned by a given sort function
- Plus some fixes in identation of resonance chamber

### DIFF
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -93,10 +93,10 @@ function Position:moveDownstairs()
             end
 
             local position = self + Position.directionOffset[direction]
-			local new_tile = Tile(position)
-			if not new_tile then return false end
+			local newTile = Tile(position)
+			if not newTile then return false end
 			
-            if new_tile:isWalkable(false, false, false, false, true) then
+            if newTile:isWalkable(false, false, false, false, true) then
                 swap(self, position)
                 return self
             end
--- a/data/lib/core/position.lua
+++ b/data/lib/core/position.lua
@@ -20,28 +20,6 @@ function Position:getNextPosition(direction, steps)
 end
 
 function Position:moveUpstairs()
-	local isWalkable = function (position)
-		local tile = Tile(position)
-		if not tile then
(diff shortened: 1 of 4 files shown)


data/lib/core/tile.lua
```diff
@@ -36,18 +36,24 @@ function Tile.relocateTo(self, toPosition)
...
+		if topStackItem and topStackItem:hasProperty(CONST_PROP_BLOCKPATH) then return false end
 	end
-
-	local items = self:getItems()
-	for i = 1, self:getItemCount() do
-		local item = items[i]
-		local itemType = item:getType()
-		if itemType:getType() ~= ITEM_TYPE_MAGICFIELD and not itemType:isMovable() and item:hasProperty(CONST_PROP_BLOCKSOLID) then
-			return false
+	if proj then
```
jlcvp, line 50: what does `proj` stands for?

andersonfaaria: projectile, if set as true it means we will check for blocksolid property in items (besides the usual check in the tile itself at the beggining)

data/lib/core/position.lua
```diff
@@ -127,14 +83,20 @@ function Position:moveDownstairs()
...
+	
+    if not tile:isWalkable(false, false, false, false, true) then
         for direction = DIRECTION_NORTH, DIRECTION_NORTHEAST do
             if direction == DIRECTION_SOUTH then
                 direction = DIRECTION_WEST
             end
 
             local position = self + Position.directionOffset[direction]
-            if isWalkable(position) then
+			local new_tile = Tile(position)
```
dudantas: Wouldn't it be interesting to switch to newTile?

dudantas (approved):
For my tests it's ok.
There is only the suggestion to change the new_tile variable to newTile

PR #1235, opened 2020-05-16
New method getDepotLocker

andersonfaaria:
There's one for chest but there isn't one for depotlocker even though we have all the mechanics of it in the source already.


lgrossi (approved):
lgtm

PR #1238, opened 2020-05-16
Monster strategies

andersonfaaria:
This PR includes all the commits below from OTHire but translated to TFS/OTX format:
- https://github.com/TwistedScorpio/OTHire/commit/8e567570eec5a2d4570a376bb7431ee280d36513
- https://github.com/TwistedScorpio/OTHire/commit/6a0528647d15ad883720015f9d22648150316e00
- https://github.com/TwistedScorpio/OTHire/commit/f8b96488a40f3f51c994f5f654a06b1fe4ad3998
- https://github.com/TwistedScorpio/OTHire/commit/7ebb966c60eaf09b97329886bdd98d16efd13ec1

Has a strong dependency on #1213 and small dependency on #1198

Improvements:
- With this ranged monsters will only dancestep when they reach ideal distance (usually 4).
- ranged monsters will change target correctly now (NEEDS #1198)  (always picking the
closest one), also all monsters have correct chance to change target and
strategies as they work in real tibia, they can pick the nearest target,
the one with the lowest % health, the one that did the most damage and
also a random one.
- Fleeing monsters don't melee attack
- With this ranged monsters will only dancestep when they reach ideal distance (usually 4).

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1284,7 +1284,7 @@ bool Monster::getDanceStep(const Position& creaturePos, Direction& moveDirection
 	uint32_t centerToDist = std::max<uint32_t>(distance_x, distance_y);
 	
 	//monsters not at targetDistance shouldn't dancestep
-	if (centerToDist < mType->info.targetDistance) {
+	if (centerToDist < (uint32_t) mType->info.targetDistance) {
 		return false;
 	}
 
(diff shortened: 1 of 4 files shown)


htc16:
Testado, pode da merge, juntamente com os monstros https://github.com/opentibiabr/otservbr-global/pull/1213.

andersonfaaria:
tested and working

PR #1259, opened 2020-05-22
Refactor Account Class

Costallat:
Refactor account class to remove static function to handle coins and to give more control over the account.

- Removes static functions
- Create a 'real' Account class to handle with things related to the account, like password, coins, players(account player type, not the Player object), adjust types, created new table to store the coins transactions of the account, not related with the store history, among other things
- Add documentation
- Add unit test
- Improve GHA filters


.editorconfig
```diff
@@ -8,10 +8,15 @@ indent_style = tab
...
 
-[*.{cpp,h,lua,xml}]
+[*.{lua,xml}]
 indent_style = tab
 indent_size = 4
 
+# Google Style
+[*.{cpp,hpp,h,sql}]
+indent_style = space
+indent_size = 2
```
lgrossi, line 18: Nice

schema.sql
```diff
@@ -23,13 +23,13 @@ INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '0'), ('mo
 --
 
 CREATE TABLE IF NOT EXISTS `accounts` (
-  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `id` int(11) UNSIGNED NOT NULL AUTO_INCREMENT,
   `name` varchar(32) NOT NULL,
   `password` char(40) NOT NULL,
   `secret` char(16) DEFAULT NULL,
   `type` int(11) NOT NULL DEFAULT '1',
   `premdays` int(11) NOT NULL DEFAULT '0',
```
lgrossi: Maybe premdays and type should be unsigned as well?

src/account.cpp
```diff
@@ -22,63 +22,417 @@
...
+ ******************************************************************************/
 
-	g_databaseTasks.addTask(query.str());
+error_t Account::LoadAccountDB() {
+  if (id_ != 0) {
+    return this->LoadAccountDB(id_);
+  } else if (!name_.empty()) {
+    return this->LoadAccountDB(name_);
+  } else {
+    LOG_F(ERROR, "Account not initialized! ID and Name without value!");
```
lgrossi: Not needed else.

src/account.cpp
```diff
@@ -22,63 +22,417 @@
...
-			"`description`, `time`) VALUES (" << accountId << "," << coins
-			<< "," << db.escapeString(description) << "," << time(nullptr)
-			<< ")";
-	DLOG_F(INFO, "Query:[%s].", query.str().c_str());
+  this->SetID(result->getNumber<uint32_t>("id"));
+  this->SetName(result->getString("name"));
+  this->SetAccountType(static_cast<AccountType>(
+                                          result->getNumber<int32_t>("type")));
+  this->SetPassword(result->getString("password"));
+  this->SetPremiumRemaningDays(result->getNumber<uint16_t>("premdays"));
```
lgrossi, line 259: Much nicer and cleaner, well done

src/account.hpp
```diff
@@ -0,0 +1,218 @@
...
+  ERROR_INVALID_ACC_TYPE,
+  ERROR_INVALID_ID,
+  ERROR_INVALID_LAST_DAY,
+  ERROR_LOADING_ACCOUNT_PLAYERS,
+  ERROR_NOT_INITIALIZED,
+  ERROR_NULLPTR,
+  ERROR_VALUE_NOT_ENOUGH_COINS,
+  ERROR_VALUE_OVERFLOW
+};
+
```
lgrossi, line 33: Maybe think about splitting errors into generic errors (db, nullptr, invalid id, no..) to be stored in enums.h and account spevific errors, to be stored here? So account specific erros can start at LAST_GENERIC_ERROR value.

src/database.h
```diff
@@ -21,8 +21,11 @@
 #define FS_DATABASE_H_A484B0CDFDE542838F506DCE3D40C693
 
 #include <boost/lexical_cast.hpp>
-
 #include <mysql/mysql.h>
+#include <memory>
+#include <mutex>
+#include <map>
+#include <iostream>
```
lgrossi, line 28: Couldnt find where those imports are being used.

src/enums.h
```diff
@@ -201,14 +203,6 @@ enum SpellType_t : uint8_t {
...
 };
 
-enum AccountType_t : uint8_t {
-	ACCOUNT_TYPE_NORMAL = 1,
-	ACCOUNT_TYPE_TUTOR = 2,
-	ACCOUNT_TYPE_SENIORTUTOR = 3,
-	ACCOUNT_TYPE_GAMEMASTER = 4,
-	ACCOUNT_TYPE_GOD = 5
-};
-
```
lgrossi, line 211: Nice shout moving it to account.

src/game.cpp
```diff
@@ -6246,69 +6307,52 @@ void Game::playerCoinTransfer(uint32_t playerId, const std::string &receiverName
...
+
+      sender_account.GetCoins(&sender_coins);
+      message.str("");
+      message << "You have successfully transfered " << amount << " coins to " << receiverName << ".";
+      sender->sendStorePurchaseSuccessful(message.str(), sender_coins);
+      if (receiver && !receiver->isOffline()) {
+        receiver->sendCoinBalance();
+      }
+    }
+  }
```
lgrossi, line 6357: Would be nice if we could add tests to those game changes as well

lgrossi (commented):
That's amazing, lgtm.
Ive added some comments/suggestions but this is a really nice refactor imho. Also, unit/integration tests is a lovely bonus feature of this PR, I really liked it.
Would be nice if you could update the readme to do a quick example of how to build and how to run tests.

Thanks for that!

schema.sql
Costallat: Premdays I left as integer because it can be used to punish some player, so it will have like '-5' premium days, so to play again he needs to buy at least 5 day to play as free player.

'type' I will check, I didn't change some fields as they are used in others table and the DB complains about the type mismatch.

src/account.cpp
Costallat: tks

src/game.cpp
Costallat: Yep, but first we need to refactor that class too, they are not prepared to do unit test very well

schema.sql
Costallat: Done, also removed others not used fields from account.

src/database.h
Costallat: map -> std::map<std::string, size_t> listNames;
memory -> using DBResult_ptr = std::shared_ptr<DBResult>;
mutex -> std::recursive_mutex databaseLock;

src/account.hpp
Costallat: Agree, I kept all things inside the class as I don't have a vision of what comes ahead, but as soon as I or another one start to refactor we can start to create those enums that can be used on other class.

src/database.h
lgrossi: I mean, I saw they being used, however they were being used before. There isnt any new addition that you did that using them. Im wandering why those imports are now needed. Im quite sute they're being imported somewhere else already, no?

lgrossi (approved):
Lgtm, minor comment about imports needs.

dudantas (approved):
LGTM

dudantas:
Thanks for your contribution!

schema.sql
```diff
@@ -23,22 +23,37 @@ INSERT INTO `server_config` (`config`, `value`) VALUES ('db_version', '0'), ('mo
 --
 
 CREATE TABLE IF NOT EXISTS `accounts` (
-	`id` int(11) NOT NULL AUTO_INCREMENT,
-	`name` varchar(32) NOT NULL,
-	`password` char(40) NOT NULL,
-	`secret` char(16) DEFAULT NULL,
```
Znote, line 29: You removed this column `secret`, which breaks Two-Factor Authentication

Costallat: @Znote I removed as I didn't find where in the code it was being used, can you show me where it is being used so I can add it back to the account? 

Tks

Znote: > @Znote I removed as I didn't find where in the code it was being used, can you show me where it is being used so I can add it back to the account?
> 
> Tks

I don't know why it is removed on OTservBR, but it is available in TFS 1.3:
https://github.com/otland/forgottenserver/blob/09befdca6ad2ce9f0d5c308705095d5693fc7fe8/src/iologindata.cpp#L120-L141

PR #1261, opened 2020-05-23
[WIP] Improvement - Item:getWeapon()

andersonfaaria:
With this PR you'll be able to extend the weapons functions from an item by simply using item:getWeapon() and then calling one of the weapons methods.

This is a chance for changing the weapons methods to allow set/get as monstertype does

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -6971,7 +6971,7 @@ int LuaScriptInterface::luaItemSerializeAttributes(lua_State* L)
 	return 1;
 }
 
-+int LuaScriptInterface::luaItemGetWeapon(lua_State* L)
+int LuaScriptInterface::luaItemGetWeapon(lua_State* L)
 {
 	// item:getWeapon()
 	Item* item = getUserdata<Item>(L, 1);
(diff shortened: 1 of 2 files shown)


src/luascript.cpp
```diff
@@ -17222,11 +17244,15 @@ int LuaScriptInterface::luaWeaponSoul(lua_State* L)
...
 	Weapon* weapon = getUserdata<Weapon>(L, 1);
 	if (weapon) {
-		weapon->setBreakChance(getNumber<uint32_t>(L, 2));
-		pushBoolean(L, true);
+		if (lua_gettop(L) == 1) {
+			lua_pushnumber(L, weapon->getBreakChance());
+		} else {
+			weapon->setBreakChance(getNumber<uint32_t>(L, 2));
+			pushBoolean(L, true);
+		}
```
lgrossi, line 17255: please, split that into getWeaponBreakChance and setWeaponBreakChance. It's much clearer and it's nice to keep a 1:1 relation between function and goal.

lgrossi (changes requested):
I strongly recommend to not use the same function to perform get/set since it can be confusing.
Otherwise, lgtm

src/luascript.cpp
andersonfaaria: I agree, but recently (specially with monstertype functions) TFS have migrated to if a function has the parameter it's a set, if not it's a get. It's really a matter of preference but we can check what makes more sense and define our own standard.

PR #1252, opened 2020-05-20
Issue Fix Pack

FakeShinoda:
- Fix #1244 (_ALESAR E YAMAN BUG_) _**Thanks to the issue creator <3**_
- Fix #1226 (_Creature Convince not showing Icons_.)
- Fix #1249 (_Missing Cobra Crossbow Attributes_) _**Thanks to the issue creator <3**_
- Fix #1225 (_OnLook bug_) _**Thanks to the issue creator <3**_

PR #1262, opened 2020-05-23
[WIP] Improvement - Creature Reload

andersonfaaria:
This PR will allow to you quickly reload creatures to all players that is inside the spectator list. Useful to avoid having to create creatures far away for updating their hp, flags, emblems, skull, etc.

Conflicts with #1252

### DIFF
--- a/src/luascript.cpp
+++ b/src/luascript.cpp
@@ -2391,6 +2391,7 @@ void LuaScriptInterface::registerFunctions()
 	registerMethod("Creature", "remove", LuaScriptInterface::luaCreatureRemove);
 	registerMethod("Creature", "teleportTo", LuaScriptInterface::luaCreatureTeleportTo);
 	registerMethod("Creature", "say", LuaScriptInterface::luaCreatureSay);
+	registerMethod("Creature", "reload", LuaScriptInterface::luaCreatureReload);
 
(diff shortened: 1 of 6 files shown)


src/luascript.cpp
```diff
@@ -8215,6 +8216,28 @@ int LuaScriptInterface::luaCreatureTeleportTo(lua_State* L)
...
+	// creature:reload()
+	Creature* creature = getUserdata<Creature>(L, 1);
+	if (!creature) {
+		lua_pushnil(L);
+		return 1;
+	}
+
+	const Position& position = creature->getPosition();
+	SpectatorHashSet spectators;
+	g_game.map.getSpectators(spectators, position, false, true); // 3 parameter is multifloor, check if there's need.
```
lgrossi, line 8230: Why not allowing multifloor as parameter?

lgrossi (commented):
nvm i didnt see the 'wip' tag.

src/luascript.cpp
andersonfaaria: can be an option, but ideally you would only need to reload a creature that is in the same floor for the designed cases (change on flag/emblem/skull/name)

andersonfaaria:
I'm closing my PR due the merge of #1252

PR #1273, opened 2020-05-26
Update quests.lua to work with Killing in the Name of...

### LINKED ISSUE #1211: Every kill Minotaur display: "Your questlog has been updated"
"Killing in the Name" 
Every kill Minotaur display: **Your questlog has been updated.**

image

YohanAugusto:
**Problem:**

Every monster killed on **Killing in the Name of...** we are getting: "Your questlog has been updated."

**Fix:**

Added a tag to avoid sending questlog updated on every storage updated. This is not applied when you start the quest, since is will not work when you get the questlog for the first time (storage value = -1)

It shoud fix #1211 

**Important:**

Remember to restart your server when you change quests' value!

### DIFF
--- a/data/lib/core/quests.lua
+++ b/data/lib/core/quests.lua
@@ -1371,6 +1371,7 @@ if not Quests then
 					storageid = 1501,
 					startvalue = 0,
 					endvalue = 100,
+					avoidMsgOnEveryUpdate = true,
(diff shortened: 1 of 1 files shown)


Lycanzito:
testado , 100 % ;)'

PR #1291, opened 2020-06-02
Resolves #1253: add lastday to the account table when purchasing premium in store

### LINKED ISSUE #1253: Casas e premium time...
Galera, seguinte....

Minha tabelas de houses está vazia... o myaac nem acha as casas.... e também verifiquei que o campo lastday do banco de dados está vazio.... mesmo comprando o premmium time, ele fica falando que a conta está freeacc...

dudantas:
Fixed the lastday update in the accounts table after buying a premium in the store.

image
image

Thanks @Heitortadeu for the report and tips on how to fix

### DIFF
--- a/data/modules/scripts/gamestore/init.lua
+++ b/data/modules/scripts/gamestore/init.lua
@@ -1087,7 +1087,7 @@ function GameStore.processAllBlessingsPurchase(player)
 end
 
 function GameStore.processPremiumPurchase(player, offerId)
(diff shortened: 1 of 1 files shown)


deletera:
Hello, I just made a test and after the days are over, the character/account still with the premdays, should it be changed something in the globalevents?

Here it's my example, the accounts with 2 and 1 premdays should be over the premdays, but they still there:

image

PR #1293, opened 2020-06-02
Small fixes in dawnport

dudantas:
Thanks for @heitortadeu by the reports!

Correction in oressa (problem when choosing vocation)
Fixed an event that teleports the player to the temple five minutes after reaching level 20)
This was a problem because, if the player left dawnport before five minutes, when the event ran, he would return to dawnport and be stuck.
Thus, I added the stopEvent on Oressa, after choosing the vocation.

Inigo was not changing the vocation to none and converting heath / mana / cap
Captain dreadnought only took to venore
The oreacle did not give a vocation for error in the variable
Small vocation_trial problem solved

Correction in migrations. 
Added another migration so that the previous one doesn't run every time you open the distro).


data/npc/scripts/oressa.lua
```diff
@@ -361,12 +362,15 @@ local function creatureSayCallback(cid, type, msg)
...
 		for index, value in pairs(topicTable)do
-			if not player:getStorageValue(Storage.Dawnport.DoorVocation) == value then
-				player:setVocation(Vocation(value))
-				player:setStorageValue(Storage.Dawnport.DoorVocation, value)
-			else
-				npcHandler.topic[cid] = 0
-				return true
+			if npcHandler.topic[cid] == index then
+				local storage = player:getStorageValue(Storage.Dawnport.DoorVocation)
+				if storage ~= 1 and storage ~= 2 and storage ~= 3 and storage ~= 4 then
```
lgrossi: Why not if storage < 1 and storage > 4? Would simplify a bit

data/npc/scripts/oressa.lua
```diff
@@ -416,6 +420,8 @@ local function creatureSayCallback(cid, type, msg)
 				"Take the ship to reach the Mainland. Farewell, friend and good luck in all you undertake!"
 			},
 		cid, false, true, 200)
+		-- For stop the event started at level 20, so that players who are already allowed to leave dawnport, do not return.
+		stopEvent(teleportToDawnportTemple)
```
lgrossi: Instead of cancelling you can just check the storage in the teleport function and do nothing if he has a vocation set or smtg like that.

data/lib/miscellaneous/dawnport.lua
```diff
@@ -181,18 +181,22 @@ function dawnportSetStats(player)
 	end
 end
 
+-- Teleport to the dawnport temple after reaching level 20 (the player has five minutes before being teleported)
+function dawnportPlayerTeleportToTemple(uid)
+	local player = Player(uid)
+	if player then
+		local town = player:getTown()
```
lgrossi: You can be safer here and validate storage values and if town = downport, so that you ensure you only teleport players that belong to dawnport.

lgrossi (commented):
Lgtm, however teleport function should be safer. Cancelling it wont ensure it wont happen. There can be a race condition between your choice and the event cancel that still teleport undesired players back to dawnport.

data/npc/scripts/oressa.lua
dudantas: Truth. I will make. Thanks!

data/lib/miscellaneous/dawnport.lua
dudantas: It doesn't work, because the player may be inside dawnport, but in a state that can no longer return to the temple, for example in the boat that takes him to main (where he can no longer return to the temple)

dudantas:
> Lgtm, however teleport function should be safer. Cancelling it wont ensure it wont happen. There can be a race condition between your choice and the event cancel that still teleport undesired players back to dawnport.

Fixed, thank you!

data/lib/miscellaneous/dawnport.lua
```diff
@@ -181,18 +181,26 @@ function dawnportSetStats(player)
 	end
 end
 
+-- Teleport to the dawnport temple after reaching level 20 (the player has five minutes before being teleported)
+function teleportToDawnportTemple(uid)
+	local player = Player(uid)
+	if player then
+		-- If not have the Oressa storage, teleport player to the temple
+		if player:getStorageValue(Storage.Dawnport.DoorVocation) == -1 then
```
lgrossi: I can't see any problem joining the conditions here, just to clean up a bit:
```
if player and player:getStorageValue(Storage.Dawnport.DoorVocation) == -1 then
```

The else and the return are also not needed, if you dont expect any return to happen.

data/scripts/creaturescripts/others/dawnport_max_level.lua
```diff
@@ -10,13 +10,8 @@ function dawnportMaxLevel.onAdvance(creature, target)
...
 			"You have reached the limit level and have to choose your vocation and leave Dawnport.")
-		addEvent(function(uid)
-		local player = Player(uid)
-			if player then
-				player:teleportTo(town:getTemplePosition())
-				player:getPosition():sendMagicEffect(CONST_ME_TELEPORT)
-			end
-		end, 5 * 60 * 1000, player:getId())
+		-- Adds the event that teleports the player to the temple in five minutes after reaching level 20
+		addEvent(teleportToDawnportTemple, 1 * 60 * 1000, player:getId())
```
lgrossi: The comment says '5 minutes' the code says '1 minute', can you please make it consistent?

schema.sql
```diff
@@ -811,9 +811,9 @@ INSERT INTO `accounts`
...
-(2, 'Druid Sample', 1, 1, 8, 2, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
-(3, 'Paladin Sample', 1, 1, 8, 3, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
-(4, 'Knight Sample', 1, 1, 8, 4, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
-(5, 'ADM', 6, 1, 2, 0, 150, 150, 0, 113, 115, 95, 39, 129, 60, 60, 1, '', 410, 1 );
+(1, 'Rook Sample', 1, 1, 2, 0, 155, 155, 100, 113, 115, 95, 39, 129, 60, 60, 1, '', 410, 0),
+(2, 'Sorcerer Sample', 1, 1, 8, 1, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
+(3, 'Druid Sample', 1, 1, 8, 2, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
+(4, 'Paladin Sample', 1, 1, 8, 3, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
+(5, 'Knight Sample', 1, 1, 8, 4, 185, 185, 4200, 113, 115, 95, 39, 129, 90, 90, 8, '', 470, 1),
+(6, 'ADM', 6, 1, 2, 0, 150, 150, 0, 113, 115, 95, 39, 129, 60, 60, 1, '', 410, 1 );
```
lgrossi, line 819: Not sure the impact of that, seems unrelated to the feature.

lgrossi (commented):
Looks nice, a few other minor adjusts (mainly clean ups) and a small inconsistency with the timing.

schema.sql
dudantas: Yes it does. They are used to create the initial dawnport char (at least the rook sample)
But as it was reported that id 0 of the problem when importing schema.sql, I had to play it forward.

data/scripts/creaturescripts/others/dawnport_max_level.lua
dudantas: Thanks, solved.

data/lib/miscellaneous/dawnport.lua
dudantas: Thanks, solved.

lgrossi: I still think that you can join those ifs in one.

lgrossi (approved):
Lgtm, ive left two minor suggestions

dudantas:
> Lgtm, ive left two minor suggestions

Thanks @lgrossi, appreciate your help!

PR #1300, opened 2020-06-04
Kick players while they are attacking or attacked by monsters (when afk)

dudantas:
Resolves: #890
All credits to @nekiro
From: https://github.com/otland/forgottenserver/pull/2990

Co-Authored-By: Marcin <nekiro@users.noreply.github.com>

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -1510,7 +1510,7 @@ void Player::setNextWalkTask(SchedulerTask* task)
 	}
 }
 
-void Player::setNextActionTask(SchedulerTask* task)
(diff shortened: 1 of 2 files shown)


omarcopires (approved):

gerotib:
I tested before and after this pr, and I see no difference. I mean I got kicked after 2 minutes (changed `kickIdlePlayerAfterMinutes` in config from 15 for test) while being attacked by monster in both cases. Only thing that I noticed related to this pr is that `idleWarningTime` and `idleKickTime` in config.lua are unused and could be removed.

dudantas:
I will bring a more suitable solution soon.

PR #1304, opened 2020-06-05
Fixing ids from sample players

dudantas:


dudantas:
Fixed in https://github.com/opentibiabr/otservbr-global/pull/1259

PR #1321, opened 2020-06-09
Fix groups "access" bug (this gave access from adm to tutors)

dudantas:
This allowed the player, if he was a tutor or senior tutor, to go through walls, have access to outfits and other things that are exclusive to group 4 or more


lgrossi (approved):
LGTM, why moving everything to lowercase? Removing access 1 wont remove some wanted access from tutor and senior as well? Do we know what is covered by access 1?

dudantas:
> LGTM, why moving everything to lowercase?

in the groups file I kept sincronized with the tfs, in the schema it is because it makes no difference, it always identifies only lowercase

PR #1372, opened 2020-06-19
Removed unsed storage

omarcopires:


lgrossi:
Very useful, and nice PR name btw. Can't decide if I go for unused or unsent. :sweat_smile:

PR #1383, opened 2020-06-23
Golden prision key script fix and Prince Drazzak loot table small fix.

nickolasdeluca:
The golden key should be removed from the player when used and Prince Drazzak should always drop a dream matter and has a 93.75 chance to drop a maximum of 3 unrealized dreams.


dudantas (approved):

dudantas:
Thanks for the contribution!

PR #1411, opened 2020-06-29
Fix player depots memory leak

lBaah:
Fix critical memory leak - player depots aren't destroyed at all 

Author: @SaiyansKing

### DIFF
--- a/src/player.cpp
+++ b/src/player.cpp
@@ -65,6 +65,12 @@ Player::~Player()
 			item->decrementReferenceCounter();
 		}
 	}
+  
+  for (const auto& it : depotChests) {
+		if (!it.second->getRealParent()) {
+			it.second->decrementReferenceCounter();
+		}
+	}
 
 	for (const auto& it : depotLockerMap) {
 		it.second->removeInbox(inbox);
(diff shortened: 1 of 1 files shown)


src/player.cpp
```diff
@@ -65,6 +65,12 @@ Player::~Player()
 			item->decrementReferenceCounter();
 		}
 	}
+  
+  for (const auto& it : depotChests) {
+		if (!it.second->getRealParent()) {
```
lgrossi, line 70: Don't we need to double check if it.second is not nullptr? I'd prefer to be redundant but safe.

lgrossi (approved):
Always good to sort out those memory leaks, tfs has plenty of them.
A small comment about the validation, might lead to a seg fault if the map has any null reference value stored under one of it's keys.

SaiyansKing:
Keep in mind that these changes might actually crash this server on Player destructor because it use multiple depot boxes that keeps getting adding to container in unsafe manner(don't increase reference counter) and might end up dereferencing deleted pointer. It would actually be best to use shared_ptr for handling these cases however it can get too much memory overhead.

lgrossi:
> Keep in mind that these changes might actually crash this server on Player destructor because it use multiple depot boxes that keeps getting adding to container in unsafe manner(don't increase reference counter) and might end up dereferencing deleted pointer. It would actually be best to use shared_ptr for handling these cases however it can get too much memory overhead.

@SaiyansKing indeed, the way we do the reference counting and the additions/removal is quite unsafe. I've recently faced an issue with creatures summons, that when a creature is deleted it't not removed from the master summons list, what causes some weird behaviors and segmentation fault. The fix was to ensure that it is removed from it's master summons array.

PR #1419, opened 2020-07-02
Boosted Creature

oualid6496:
To Do:
- [x] Roll Boosted Monster
- [x] Boosted monster shouldn't change if restart happens on the same day.
- [x] Boosted monster should change if restart happens on another day.
- [x] Boosted monster should get double exp.
- [x] Boosted monster should get double loot.
- [ ] Boosted monster should get double respawn.
- [x] There should be a output json file (to work with the client).
- [ ] On login there should be a message displaying which monster is boosted today.

Info:
- Path for the Json file needs to be configured in the lib file.
- List of possible boosted creatures can also be configured in the lib file (the list of bestiary is being used now)


oualid6496:
if anyone wants to finish this feel free to do so :)

majestyotbr:
@gpedro PR waiting for your review.

data/lib/others/boostedCreature.lua
```diff
@@ -0,0 +1,672 @@
+-- Made by Nekiro
+if not BoostedCreature then
+	BoostedCreature = {name = ""}
+end
+
+BoostedCreature.jsonPath = "D:/xampp/htdocs/boostedCreature.json"
```
gpedro, line 6: this should be fetched by database

data/lib/others/boostedCreature.lua
```diff
@@ -0,0 +1,672 @@
...
+BoostedCreature.storages = {race = 99999, timestamp = 99998} -- make sure they arent the same and change to whatever you like
+
+BoostedCreature.races = {
```
gpedro, line 9: can we reuse bestiary races?

majestyotbr:
@marcosvf132 did a new PR for Boosted Creature. This can be closed.

PR #1420, opened 2020-07-02
Fixed Grizzly Adams not buying anything from the player.

nickolasdeluca:
Grizzly wasn't buying anything the player tried to sell and was throwing an error.
I was able to fix this issue but this script probably needs a remake, It's messy and probably outdated, I am currently not skilled enough to do it but this at least makes him able to trade again.

### DIFF
--- a/data/npc/scripts/grizzly_adams.lua
+++ b/data/npc/scripts/grizzly_adams.lua
@@ -132,9 +132,8 @@ local function onBuy(cid, item, subType, amount, ignoreCap, inBackpacks)
 	if not ignoreCap and player:getFreeCapacity() < ItemType(items[item].id):getWeight(amount) then
 		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'You don\'t have enough cap.')
 	end
-	if not player:removeMoneyNpc(items[item].buy * amount) then
-		selfSay("You don't have enough money.", cid)
-	else
+	if items[item].buy then
+		player:removeMoneyNpc(amount * items[item].buy)
 		player:addItem(items[item].id, amount)
 		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'Bought '..amount..'x '..items[item].name..' for '..items[item].buy * amount..' gold coins.')
 	end
@@ -143,12 +142,10 @@ end
 
 local function onSell(cid, item, subType, amount, ignoreCap, inBackpacks)
 	local player = Player(cid)
-	local items = setNewTradeTable(getTable(player))
-	if items[item].sellPrice and player:removeItem(items[item].itemId, amount) then
-		player:addMoney(items[item].sellPrice * amount)
-		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'Sold '..amount..'x '..items[item].realName..' for '..items[item].sellPrice * amount..' gold coins.')
-	else
-		selfSay("You don't have item to sell.", cid)
+	if items[item].sell then
+		player:addMoney(items[item].sell * amount)
+		player:removeItem(items[item].id, amount)
+		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'Sold '..amount..'x '..items[item].name..' for '..items[item].sell * amount..' gold coins.')
 	end
 	return true
 end
--- a/data/npc/scripts/grizzly_adams.lua
+++ b/data/npc/scripts/grizzly_adams.lua
@@ -1,4 +1,4 @@
-	local keywordHandler = KeywordHandler:new()
+local keywordHandler = KeywordHandler:new()
 local npcHandler = NpcHandler:new(keywordHandler)
 NpcSystem.parseParameters(npcHandler)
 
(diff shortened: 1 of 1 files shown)


data/npc/scripts/grizzly_adams.lua
```diff
@@ -132,9 +132,8 @@ local function onBuy(cid, item, subType, amount, ignoreCap, inBackpacks)
 	if not ignoreCap and player:getFreeCapacity() < ItemType(items[item].id):getWeight(amount) then
 		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'You don\'t have enough cap.')
 	end
-	if not player:removeMoneyNpc(items[item].buy * amount) then
-		selfSay("You don't have enough money.", cid)
-	else
+	if items[item].buy then
+		player:removeMoneyNpc(amount * items[item].buy)
```
gpedro: you need to check if the money is removed

data/npc/scripts/grizzly_adams.lua
```diff
@@ -143,12 +142,10 @@ end
...
 	local player = Player(cid)
-	local items = setNewTradeTable(getTable(player))
-	if items[item].sellPrice and player:removeItem(items[item].itemId, amount) then
-		player:addMoney(items[item].sellPrice * amount)
-		return player:sendTextMessage(MESSAGE_INFO_DESCR, 'Sold '..amount..'x '..items[item].realName..' for '..items[item].sellPrice * amount..' gold coins.')
-	else
-		selfSay("You don't have item to sell.", cid)
+	if items[item].sell then
+		player:addMoney(items[item].sell * amount)
+		player:removeItem(items[item].id, amount)
```
gpedro: you need to check if the item is removed before addMoney

gpedro (changes requested):

oalbsilva:
have another problem in this NPC:
If you said trade X times to NPC, it is duplicated items on trade window. 

Need check npcHandler.topic[cid] on msgcontains('trade', msg)

nickolasdeluca:
> have another problem in this NPC:
> If you said trade X times to NPC, it is duplicated items on trade window.
> 
> Need check npcHandler.topic[cid] on msgcontains('trade', msg)

Yep, that's why I said this npc needs a complete rework, fixing these issues, I was merely getting it back to work. I tried to fix some issues with his dialogues but I couldn't, I'm still learning lua and how npc dialogue works.

I'll fix the issues gpedro specified but I believe someone more experienced should recreate Grizzly's lua entirely.

oalbsilva:
> > have another problem in this NPC:
> > If you said trade X times to NPC, it is duplicated items on trade window.
> > Need check npcHandler.topic[cid] on msgcontains('trade', msg)
> 
> Yep, that's why I said this npc needs a complete rework, fixing these issues, I was merely getting it back to work. I tried to fix some issues with his dialogues but I couldn't, I'm still learning lua and how npc dialogue works.
> 
> I'll fix the issues gpedro specified but I believe someone more experienced should recreate Grizzly's lua entirely.

I don't think need recreate entirely...
just put a condition check if player already said trade to NPC. Something like this:

```lua
...
if msgcontains('trade', msg) then
    if npcHandler.topic[cid] == 4 then
        return false;
    end
    npcHandler.topic[cid] = 4
...
```

nickolasdeluca:
> I don't think need recreate entirely...
> just put a condition check if player already said trade to NPC. Something like this:
> 
> ```lua
> ...
> if msgcontains('trade', msg) then
>     if npcHandler.topic[cid] == 4 then
>         return false;
>     end
>     npcHandler.topic[cid] = 4
> ...
> ```

k, I'll give it a go, thanks for the code.

data/npc/scripts/grizzly_adams.lua
nickolasdeluca: Added the required validations.

data/npc/scripts/grizzly_adams.lua
nickolasdeluca: Added the required validations.

nickolasdeluca:
> 
> I don't think need recreate entirely...
> just put a condition check if player already said trade to NPC. Something like this:
> 
> ```lua
> ...
> if msgcontains('trade', msg) then
>     if npcHandler.topic[cid] == 4 then
>         return false;
>     end
>     npcHandler.topic[cid] = 4
> ...
> ```

Turns out the issue wasn't in the dialogue but rather on the joinTables function.

dudantas (commented):
It looks ok, only the warnings need to be resolved.

nickolasdeluca:
> It looks ok, only the warnings need to be resolved.

Well, that's the best I can do, If I fix the lines bigger than 120 characters then luacheck complains that the line has trailing white spaces. Dunno what it want's me to do. I could create a local var to store the messages but that feels unnecessary.

nickolasdeluca:
> It looks ok, only the warnings need to be resolved.

Done, fixed all the warnings.

dudantas (approved):

PR #1424, opened 2020-07-03
Heart of Destruction Creaturescripts review

lBaah:
Remove unused variable from aftershock_transform
Remove unused variable from anomaly_transform
Remove unused variable from charging_out_death
Move creature:remove() to the end of the file on crackler_transform and depolarized_transform
Remove unused variable from eradicator_transform
Remove unused variable from foreshock_transform
Change few checks at heart_boss_death
Check if creature is a monster before taking and process its name at heart_minion_death
Remove unused variable from outburst_charge
Change few checks at overcharge_death
Hardcoded magic effect at rupture_heal
Check if it's monster at rupture_resonance
Remove unused variables at shocks_death

### DIFF
--- a/data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
+++ b/data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
@@ -12,7 +12,7 @@ local function setStorage()
 					local creatures = tile:getCreatures()
 					if creatures and #creatures > 0 then
 						for _, creature in pairs(creatures) do
-							if creature:isPlayer() and creature:getStorageValue(14320) < 1then
+							if creature:isPlayer() and creature:getStorageValue(14320) < 1 then
 								creature:setStorageValue(14320, 1) -- Access to boss Anomaly
 							end
 						end
--- a/data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
+++ b/data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
@@ -11,11 +11,9 @@ local function setStorage()
 				if tile then
 					local creatures = tile:getCreatures()
 					if creatures and #creatures > 0 then
-						for _, c in pairs(creatures) do
-							if isPlayer(c) then
-								if c:getStorageValue(14320) < 1 then
-									c:setStorageValue(14320, 1) -- Access to boss Anomaly
-								end
+						for _, creature in pairs(creatures) do
+							if creature:isPlayer() and creature:getStorageValue(14320) < 1then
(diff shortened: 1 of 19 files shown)


majestyotbr:
@oualid6496 Can you test? Thanks.

oualid6496:
The crackler bug has been resolved with this. Tested that yesterday. The rest I can test later today.

data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
```diff
@@ -11,11 +11,9 @@ local function setStorage()
 				if tile then
 					local creatures = tile:getCreatures()
 					if creatures and #creatures > 0 then
-						for _, c in pairs(creatures) do
-							if isPlayer(c) then
-								if c:getStorageValue(14320) < 1 then
-									c:setStorageValue(14320, 1) -- Access to boss Anomaly
-								end
+						for _, creature in pairs(creatures) do
+							if creature:isPlayer() and creature:getStorageValue(14320) < 1then
```
oualid6496: There needs to be a space between 1 and then

oualid6496 (approved):
Everything (creaturescripts related) seems to be working fine except for the typo in overcharge_death,
There are some problems with the vortexes that don't teleport but that doesn't have anything to do with the changes in this PR.

Look at the comment that I made in overcharge_death.lua and ty for solving it!

data/creaturescripts/scripts/quests/heart_of_destruction/overcharge_death.lua
lBaah: Thank you!

oualid6496 (approved):
Vortexes are working again, ty!
(I've only tested one but they're all using the same system so the others are most likely to also work)

oualid6496:
@lBaah could you add 1 more change? In monster/otherworld/unstable_spark.xml
`<event name="SparksDeath" />`

should be changed into
`<event name="SparkDeath" />`

or else it won't register it's death and you won't be able to get acces to the boss.

lBaah:
Is this monster spawned with those scripts? If the monster are created with Game.createMonster then we can just set the event to that monster and unregister it when it dies.

Edit:
https://github.com/opentibiabr/otservbr-global/blob/4a1bf5f21aabe6286442aa8ade4f77318763524a/data/actions/scripts/quests/heart_of_destruction/sparks_lever.lua#L101

looks like it is spawning from here

oualid6496:
Yea I guess it could be done like that. Both ways are ok since these monsters are only being used at that part of the quest.

majestyotbr (approved):

PR #1438, opened 2020-07-07
Fix possible crash on addGuild

lBaah:
I've managed to cast a powerful spell on this code, so, from now on, it will check if the pointer point to something before trying to add it. Do you get the point?
Extremely powerful magics happening!


gpedro:
approved. 

ps: this already fixed on stuck PR #1105

gpedro (approved):

Issue #1445, opened 2020-07-10
PVP without activating.


ivansho92:
in this last branch, if players stay in party, can make damage at members party, if you do some spell of attack, don't care if you opened or closed your hand.
In this example, these two players are in party and make damage with spells.
http://imgfz.com/i/8pVmPLo.png

dudantas:
the server is retro-pvp, this is normal, isn't it?

Tazer89:
> in this last branch, if players stay in party, can make damage at members party, if you do some spell of attack, don't care if you opened or closed your hand.
> In this example, these two players are in party and make damage with spells.
> http://imgfz.com/i/8pVmPLo.png

Enable it shall work i guess? In events/events.xml

```
<event class="Creature" method="onTargetCombat" enabled="0" />
<!-- removeCombatProtection on secure modes protection -->
```

ivansho92:
> > in this last branch, if players stay in party, can make damage at members party, if you do some spell of attack, don't care if you opened or closed your hand.
> > In this example, these two players are in party and make damage with spells.
> > http://imgfz.com/i/8pVmPLo.png
> 
> Enable it shall work i guess? In events/events.xml
> 
> ```
> <event class="Creature" method="onTargetCombat" enabled="0" />
> 
> ```

yeah, fix with that xd

dudztroyer:
Normal behaviour on retro-pvp

PR #1446, opened 2020-07-10
Adding worldType "retro-pvp" by default

dudantas:
This will put an end to some complaints regarding the type of world, which by default will now be retro-pvp.
Until we have the other types of worlds working normally, since it is very difficult to maintain compatibility between all types of world.

This related to the #1445


majestyotbr:
Need test.

gabz182:
Tested.
Working, Thanks!

majestyotbr (approved):

PR #1455, opened 2020-07-13
Fix rate spawn

### LINKED ISSUE #1454: Rate spawn in config.lua don't works.
**Describe the bug**
When I change the ratespawn in config.lua nothing happens.

**Steps to reproduce the behavior:**
1. First time I change the ratespawn for 10x, and test it, comparing with normal rate.
2. Second time, I change the ratepsawn for 20x, and test it again 
3. The time that respawn born is the same in first and second test comparing with normal rate.

**Expected behavior**
I expected that respawn to be more fast than normal rate.

I'm in doubt if this is a bug or there is another place to make this change.

oalbsilva:
Fix #1454

After this PR, rateSpawn in config.lua let it work properly.

For example, a monster appears every 90 seconds, when rateSpawn = 2, it will appear every 45 seconds

### DIFF
--- a/src/spawn.cpp
+++ b/src/spawn.cpp
@@ -91,7 +91,7 @@ bool Spawns::loadFromXml(const std::string& fromFilename)
 					centerPos.y + pugi::cast<uint16_t>(childNode.attribute("y").value()),
 					centerPos.z
 				);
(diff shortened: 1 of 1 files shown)


gabz182:
Tested.
Working, thanks!

dudantas (approved):

PR #1493, opened 2020-07-18
Update world

coney4:
added missing rope spots in thais rat cave


majestyotbr:
We don't accept PRs with changes on the map, because in the past some people did proposital things. If you want to contribute, send the fixed part to @dudantas in our project discord: https://discordapp.com/invite/3NxYnyV

coney4:
Ok I understand, but how i send fixed part of map ? Should i write some coords of added tiles or just some part of map or what ?

dudantas:
> Ok I understand, but how i send fixed part of map ? Should i write some coords of added tiles or just some part of map or what ?

you can just cut out the modified area, send it zipped and tell me the position it is in ... so I paste it on the map and go up

coney4:
ok ty i will try it later

coney4:
Last question : position have to be same or it doesnt matter? Ex. i cut something in 32000,32000,8 and copy it to new map to same position or i can copy it in 1000,1000,8

dudantas:
> Last question : position have to be same or it doesnt matter? Ex. i cut something in 32000,32000,8 and copy it to new map to same position or i can copy it in 1000,1000,8

You can put it in any position, just write the original position of the area and send it to me (it could be the position of the center of the area that you cut out, for example)

PR #1508, opened 2020-07-19
fix monster death

### LINKED ISSUE #1464: Monster do not die instantly
Monsters do stay in place for about 0.5 sec before they die and transforms into corpse.

I found a fix for it to make it work like on real tibia.
https://github.com/otland/forgottenserver/pull/3030/files/8f748e3bf55e6b4f41f52b217dea652e6212203b#diff-d692aa92dfdf2ae98d790503b68ad865
I tested it and it's working without problems.

coney4:
fix #1464
I tested it killed some monsters and it works.


gpedro:
cherry-picked from otland/forgottenserver#3030

majestyotbr (approved):

PR #1528, opened 2020-07-21
Fix bestiary point count

oalbsilva:
Before: The bestiary point only go to most damage player.
After: points for everyone who did damage

### DIFF
--- a/data/creaturescripts/scripts/others/bestiary_kill.lua
+++ b/data/creaturescripts/scripts/others/bestiary_kill.lua
@@ -2,7 +2,20 @@ function onKill(player, creature, lastHit)
 	if not player:isPlayer() or not creature:isMonster() or creature:getMaster() or creature:isPlayer() then
 		return true
 	end
-	player:addBestiaryKill(creature:getName())
 	
+	local players = {}
+	for playerid, damage in pairs(creature:getDamageMap()) do
+		if (not players[playerid]) then
+			players[playerid] = true
+		end
+	end
+
+	for id,v in pairs(players) do
+		local player = Player(id)
+		if player then
+			player:addBestiaryKill(creature:getName())
+		end
+	end
+
 	return true
 end
--- a/data/creaturescripts/scripts/others/bestiary_kill.lua
+++ b/data/creaturescripts/scripts/others/bestiary_kill.lua
@@ -2,20 +2,13 @@ function onKill(player, creature, lastHit)
 	if not player:isPlayer() or not creature:isMonster() or creature:getMaster() or creature:isPlayer() then
 		return true
 	end
-	
-	local players = {}
-	for playerid, damage in pairs(creature:getDamageMap()) do
-		if (not players[playerid]) then
-			players[playerid] = true
-		end
-	end
 
-	for id,v in pairs(players) do
-		local player = Player(id)
-		if player then
-			player:addBestiaryKill(creature:getName())
+	for cid, damage in pairs(creature:getDamageMap()) do
+		local player = Player(cid)
+		if player and player:isPlayer() then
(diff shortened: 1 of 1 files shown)


dudztroyer:
Bestiary only goes to the player who killed, in global. Not whole party... AFAIK

aurelion5670:
On real tibia it works to every player that does damage to the monster

aurelion5670 (approved):

data/creaturescripts/scripts/others/bestiary_kill.lua
```diff
@@ -2,7 +2,20 @@ function onKill(player, creature, lastHit)
 	if not player:isPlayer() or not creature:isMonster() or creature:getMaster() or creature:isPlayer() then
 		return true
 	end
-	player:addBestiaryKill(creature:getName())
 	
+	local players = {}
+	for playerid, damage in pairs(creature:getDamageMap()) do
```
FakeShinoda: Why 2 loop? you can do all you need in the damageMap loop

FakeShinoda (changes requested):

data/creaturescripts/scripts/others/bestiary_kill.lua
gpedro: this can be replaced by
```lua
for cid, damage in pairs(creature:getDamageMap()) do
  local player = Player(cid)
  if player and player:isPlayer() then
    player:addBestiaryKill(creature:getName())
  end
end
```

attack787:
Working perfectly like tibia  global. thanks for job :D

data/creaturescripts/scripts/others/bestiary_kill.lua
oalbsilva: creature:getDamageMap can return duplicate ids players.
this occurs for players who can deal different types of damage to a creature

So, two loops to ensure that we don't have duplicate players

gpedro: are you sure?
https://github.com/opentibiabr/otservbr-global/blob/develop/src/creature.cpp#L1001-L1021

oalbsilva: > are you sure?
> https://github.com/opentibiabr/otservbr-global/blob/develop/src/creature.cpp#L1001-L1021

:dancers: 
I will check, because when i put only one loop the points is duplicated :-1:

oalbsilva: resolved https://github.com/opentibiabr/otservbr-global/pull/1528/commits/51ffa6c583b2221f56683d4ae2151553a75e68fc

data/creaturescripts/scripts/others/bestiary_kill.lua
```diff
@@ -2,7 +2,13 @@ function onKill(player, creature, lastHit)
 	if not player:isPlayer() or not creature:isMonster() or creature:getMaster() or creature:isPlayer() then
 		return true
 	end
-	player:addBestiaryKill(creature:getName())
-	
+
+	for cid, damage in pairs(creature:getDamageMap()) do
+		local player = Player(cid)
```
majestyotbr: Check warning on line 7 in data/creaturescripts/scripts/others/bestiary_kill.lua


GitHub Actions
/ reviewers

data/creaturescripts/scripts/others/bestiary_kill.lua#L7
[luacheck] reported by reviewdog 🐶
shadowing definition of argument 'player' on line 1

Raw Output:
data/creaturescripts/scripts/others/bestiary_kill.lua:7:9: shadowing definition of argument 'player' on line 1

majestyotbr (approved):

PR #1563, opened 2020-07-26
Gamestore update

Geekbuddy:
Change offertype training to charges instead because that's basically is what it is.
Fixed magic gold converter with the charges change.
Added maximum check for prey wild cards, reward tokens in purchase process.


majestyotbr:
And the Gold Converter?

Geekbuddy:
> And the Gold Converter?

Fixed 👍

MajestyOTBR:
The exercise weapons, gold converter and magic gold converted stopped working with the changes, you will need to revert.
Can you revert the code about guild system? So i can accept the other PR created after, it's not necessary to do this.

Geekbuddy:
> The exercise weapons, gold converter and magic gold converted stopped working with the changes, you will need to revert.
> Can you revert the code about guild system? So i can accept the other PR created after, it's not necessary to do this.

Its working becouse its tested before posting, needs to add both init, and gamestore.

MajestyOTBR:
It's working, probably i forgot something.

majestyotbr (approved):

PR #1571, opened 2020-07-26
Gamestore fixes

### LINKED ISSUE #1568: Hirelings services available even if player has no one
## CAUTION: OUR SERVER IS UPDATED DAILY. MAKE SURE YOU ARE REPORTING SOMETHING ON THE LATEST VERSION OF THE SERVER.

**Describe the bug**
Hirelings services on gamestore, like name change, sex change, dresses and abilities are available to be bought even if the player has no hirelings at all.

**To Reproduce**
0. Make sure your test player does not have bought a hireling for its house yet
1. Go to gamestore
2. Go to "Hirelings" and/or "Hirelings Dresses"
3. See that options related to hirelings are available (instead of being available just hirelings themselves to be bought at first)
4. Buy something (like a name change)
5. See the error
6. Also note the same problem informed on #1566 happens here aswell (in case of buying a hireling name change without owning a hireling player will be charged 2 times for it).

**Expected behavior**
Enable these offers just when player has at least one hireling (and for 6: be charged just once and only if choosing the hireling and succesfully changing its name).

 - OS: Windows
 - Version: 10

Geekbuddy:
Added disable if not have an hireled for outfits, skill of hirelings also if you have 10 hireled.


data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
 				disabled = 1
 				disabledReason = "You already have an active XP boost."
 			end
+		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING then
+			if self:getHirelingsCount() == 10 then
```
ghost:
```suggestion
			if self:getHirelingsCount() >= 10 then
```
Even if it isn't expected to happen, isn't better to check if it's equal or higher than 10?
So even if somehow player manages to get an hypothetical eleventh Hireling the code will keep working.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+				disabled = 1
+				disabledReason = "You have maximum of 10 hirelings."
```
ghost:
```suggestion
				disabledReason = "You already have bought the maximum number of allowed hirelings."
```
This way, if in the future it's implemented a configuration so the server owner can choose how much hirelings the players can have it will still be correct.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+			end
 		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_SKILL then
 			local skill = (HIRELING_STORAGE.SKILL + offer.id)
 			if self:hasHirelingSkill(skill) then
 				disabled = 1
 				disabledReason = "This skill is already unlocked."
 			end
+			if self:getHirelingsCount() == 0 then
+				disabled = 1
+				disabledReason = "You need to have an hireling."
```
ghost:
```suggestion
				disabledReason = "You need to have a hireling."
```
Small fix: `a` instead of `an`.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+			end
 		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_OUTFIT then
 			local outfit = offer.id - HIRELING_STORAGE.OUTFIT
 			if self:hasHirelingOutfit(outfit) then
 				disabled = 1
 				disabledReason = "This hireling outfit is already unlocked."
 			end
+			if self:getHirelingsCount() == 0 then
+				disabled = 1
+				disabledReason = "You need to have an hireling."
```
ghost:
```suggestion
				disabledReason = "You need to have a hireling."
```
Small fix: `a` instead of `an`.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+			end
+		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_NAMECHANGE then
+			if self:getHirelingsCount() == 0 then
+				disabled = 1
+				disabledReason = "You need to have an hireling."
```
ghost:
```suggestion
				disabledReason = "You need to have a hireling."
```
Small fix: `a` instead of `an`.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+			end
+		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_SEXCHANGE then
+			if self:getHirelingsCount() == 0 then
+				disabled = 1
+				disabledReason = "You need to have an hireling."
```
ghost:
```suggestion
				disabledReason = "You need to have a hireling."
```
Small fix: `a` instead of `an`.

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+				disabled = 1
+				disabledReason = "You have maximum of 10 hirelings."
+			end
 		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_SKILL then
 			local skill = (HIRELING_STORAGE.SKILL + offer.id)
 			if self:hasHirelingSkill(skill) then
 				disabled = 1
 				disabledReason = "This skill is already unlocked."
 			end
+			if self:getHirelingsCount() == 0 then
```
ghost:
```suggestion
			if self:getHirelingsCount() <= 0 then
```
Maybe I'm a bit too cautious, but isn't it better to check it like this instead?

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+				disabled = 1
+				disabledReason = "You need to have an hireling."
+			end
 		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_OUTFIT then
 			local outfit = offer.id - HIRELING_STORAGE.OUTFIT
 			if self:hasHirelingOutfit(outfit) then
 				disabled = 1
 				disabledReason = "This hireling outfit is already unlocked."
 			end
+			if self:getHirelingsCount() == 0 then
```
ghost:
```suggestion
			if self:getHirelingsCount() <= 0 then
```
Maybe I'm a bit too cautious, but isn't it better to check it like this instead?

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+				disabled = 1
+				disabledReason = "You need to have an hireling."
+			end
+		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_NAMECHANGE then
+			if self:getHirelingsCount() == 0 then
```
ghost:
```suggestion
			if self:getHirelingsCount() <= 0 then
```
Maybe I'm a bit too cautious, but isn't it better to check it like this instead?

data/modules/scripts/gamestore/init.lua
```diff
@@ -560,18 +560,41 @@ function Player.canBuyOffer(self, offer)
...
+				disabled = 1
+				disabledReason = "You need to have an hireling."
+			end
+		elseif offer.type == GameStore.OfferTypes.OFFER_TYPE_HIRELING_SEXCHANGE then
+			if self:getHirelingsCount() == 0 then
```
ghost:
```suggestion
			if self:getHirelingsCount() <= 0 then
```
Maybe I'm a bit too cautious, but isn't it better to check it like this instead?

majestyotbr (approved):

PR #1572, opened 2020-07-26
Character World Transfer

Geekbuddy:
Fix/added Storage.ExpBoost and Storage.ExpBoostCooldown.
Added Character World Transfer and Express.

If you have any ide to improve it let us know.

How it works:

WorldTransfer = 30054, ------ Mark that world transfer is active and paid.
WorldTransferCooldown = 30055,  ---- Cooldown 6 months as normal Character World Transfer.
WorldTransferId = 30056,   ---- Adds value that will be the world_id that will be grabbed of npc sharon when he change the world_id.

Worlds can be added in inti.lua but should find an other way to grab it from worlds.xml in feature? ID is world_id

	local gameworlds = {
	    [1] = {id = "1", name = "Lordaeron PVP"},
	    [2] = {id = "2", name = "Arthas PVE"},
	}


data/globalevents/scripts/others/startup.lua
```diff
@@ -63,7 +63,7 @@ function onStartup()
 	db.asyncQuery('TRUNCATE TABLE `players_online`')
 
 	-- reset storages and allow purchase of boost in the store
-	db.query('UPDATE `player_storage` SET `value` = 0 WHERE `player_storage`.`key` = 51052')
```
Geekbuddy, line 66: Changed becouse match better into the Storages.lua table i mean the storage key.

majestyotbr:
This PR will need to be reviewed carefully because it depends of the multiworld system.

dudantas:
Hello. Since there are many issues to be solved, many systems to finish and new things to be done, we will close the pull, as it is very unlikely that he will go to develop. But, we are very happy with your contribution and in the future it may be reopened and reused. We hope your understanding.

PR #1596, opened 2020-07-29
Fix: resolves #1589

### LINKED ISSUE #149: Blessing charms doesn't work
Lua Script Error: [Action Interface]
data/actions/scripts/others/blessing_charms.lua:onUse
data/modules/scripts/blessings/blessings.lua:206: attempt to index field 'Charms' (a nil value)
stack traceback:
        [C]: in function '__index'
        data/modules/scripts/blessings/blessings.lua:206: in function <data/modules/scripts/blessings/blessings.lua:205>

### LINKED ISSUE #1589: Blessing charms doesn't work
Lua Script Error: [Action Interface]
data/actions/scripts/others/blessing_charms.lua:onUse
data/modules/scripts/blessings/blessings.lua:206: attempt to index field 'Charms' (a nil value)
stack traceback:
        [C]: in function '__index'
        data/modules/scripts/blessings/blessings.lua:206: in function <data/modules/scripts/blessings/blessings.lua:205>

marcosvf132:
Resolves #1589

### DIFF
--- a/data/modules/scripts/blessings/blessings.lua
+++ b/data/modules/scripts/blessings/blessings.lua
@@ -205,18 +205,17 @@ end
 Blessings.useCharm = function(player, item)
 	for index, value in pairs(Blessings.All) do
 		if item.itemid == value.charm then
-			local useItem = value
-	if not useItem then
+	if not value then
 		return true
 	end
 
-	if player:hasBlessing(useItem.id) then
+	if player:hasBlessing(value.id) then
 		player:say('You already possess this blessing.', TALKTYPE_MONSTER_SAY)
 		return true
 	end
 
-	player:addBlessing(useItem.id, 1)
-	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, useItem.name .. ' protects you.')
+	player:addBlessing(value.id, 1)
+	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, value.name .. ' protects you.')
 	player:getPosition():sendMagicEffect(CONST_ME_LOSEENERGY)
 	item:remove(1)
 	return true
--- a/data/modules/scripts/blessings/blessings.lua
+++ b/data/modules/scripts/blessings/blessings.lua
@@ -203,21 +203,25 @@ Blessings.getPvpBlessingCost = function(level)
 end
 
 Blessings.useCharm = function(player, item)
-	local useItem = Blessings.Charms[item.itemid]
+	for index, value in pairs(Blessings.All) do
+		if item.itemid == value.charm then
+			local useItem = value
 	if not useItem then
 		return true
 	end
 
-	if player:hasBlessing(useItem.blessId) then
+	if player:hasBlessing(useItem.id) then
 		player:say('You already possess this blessing.', TALKTYPE_MONSTER_SAY)
 		return true
 	end
 
-	player:addBlessing(useItem.blessId, 1)
-	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, useItem.text .. ' protects you.')
+	player:addBlessing(useItem.id, 1)
+	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, useItem.name .. ' protects you.')
 	player:getPosition():sendMagicEffect(CONST_ME_LOSEENERGY)
 	item:remove(1)
 	return true
+		end
+	end
 end
 
 Blessings.checkBless = function(player)


data/modules/scripts/blessings/blessings.lua
```diff
@@ -203,21 +203,25 @@ Blessings.getPvpBlessingCost = function(level)
 end
 
 Blessings.useCharm = function(player, item)
-	local useItem = Blessings.Charms[item.itemid]
+	for index, value in pairs(Blessings.All) do
+		if item.itemid == value.charm then
+			local useItem = value
```
ghost: Small fix:
```suggestion
	for index, value in pairs(Blessings.All) do
		if item.itemid == value.charm then
			local useItem = value
		end
	end
```

data/modules/scripts/blessings/blessings.lua
```diff
@@ -203,21 +203,25 @@ Blessings.getPvpBlessingCost = function(level)
...
 
-	player:addBlessing(useItem.blessId, 1)
-	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, useItem.text .. ' protects you.')
+	player:addBlessing(useItem.id, 1)
+	player:sendTextMessage(MESSAGE_EVENT_ADVANCE, useItem.name .. ' protects you.')
 	player:getPosition():sendMagicEffect(CONST_ME_LOSEENERGY)
 	item:remove(1)
 	return true
+		end
+	end
```
ghost, line 223: Small fix:
```suggestion
```

ghost (changes requested):
Aside from the two fixes that I mentioned I think it is good to go.

marcosvf132:
I removed `local useItem = value` from the commit, I'm using only `value` now, there was no reasing to keep this one.

data/modules/scripts/blessings/blessings.lua
```diff
@@ -203,21 +203,24 @@ Blessings.getPvpBlessingCost = function(level)
 end
 
 Blessings.useCharm = function(player, item)
-	local useItem = Blessings.Charms[item.itemid]
-	if not useItem then
+	for index, value in pairs(Blessings.All) do
+		if item.itemid == value.charm then
```
ghost, line 207:
```lua
	for index, value in pairs(Blessings.All) do
		if item.itemid == value.charm then
```
There's no `end` closing this `for` and this `if` isn't doing anything (also there's no closing to this statement as well). I think Lua doesn't work without proper starting/ending block codes, even if they have just one line, unlike C for example.

No error was caught because on lines 222-223 there are extra `end` that will close those statements.

data/modules/scripts/blessings/blessings.lua
```diff
@@ -203,21 +203,24 @@ Blessings.getPvpBlessingCost = function(level)
...
+	if not value then
```
ghost, line 208: `value` exists just within the `for` scope, so when the `for` statement is correctly closed it'll generate an error.

data/modules/scripts/blessings/blessings.lua
ghost: My comment about those lines is still valid.

ghost (changes requested):
If I'm not misunderstanding, the new code probably won't work.
I added some comments explaining my point.

marcosvf132:
The `end` on the lines **222**-**223** was intented to close `for` (line 206) and `if` (Line 207), thats why I'm using value on the other parts of the script,  when the script find the right **item.itemid** on the line 207 it will be properly closed at the `return` on line 221, On my test server, the script is working as expected. I will keep testing it but I can't see why it won't work.

ghost:
You're right, my bad.
I just had to fix indentation to understand your code corretly.

Still, now that I read the code I think there's a easier solution.

marcosvf132:
Using the `local useItem = value` solution, removing the both ends and use useitem on the rest of the scripts just like you said is a good idea tho, maybe it can be 'lighter' then my PR, if you open a solution for this issuse, i will be glad to test it too.

ghost:
Actually I think this could be fixed just changing this line on original code:
`local useItem = Blessings.Charms[item.itemid]`

To get its value from `Blessings.All` (it has a "charm" index with item id).

Edit: nevermind, the original code uses lots of wrong indexes. Your solution is the best possible (and I think I need to sleep since I'm taking too much time to interpret a code :P).

majestyotbr (approved):

PR #1600, opened 2020-07-30
Removed outdated check on combat.cpp

ghost:
Removed outdated check that disables combat with no vocation characters.

There are enough better ways to deal with it today like using non-PVP tiles on map or even using `protectionLevel` option on config.lua.

### DIFF
--- a/src/combat.cpp
+++ b/src/combat.cpp
@@ -279,10 +279,6 @@ bool Combat::isProtected(const Player* attacker, const Player* target)
 		return true;
 	}
 
-	if (attacker->getVocationId() == VOCATION_NONE || target->getVocationId() == VOCATION_NONE) {
-		return true;
-	}
-
 	if (attacker->getSkull() == SKULL_BLACK && attacker->getSkullClient(target) == SKULL_NONE) {
 		return true;
 	}


andersonfaaria:
Hi @kairion, I've been seeing the discussion on otland PR and I wanted to add my 2cents here before approving the merge:
From my understand, protection level would mitigate this but after the people going over the protection level, if we remove this limitation of not able to pvp being 'no vocation' that would mean that people could simply PVP in rookgard (unless we turn the whole map as No-PVP, which is more work than simply having this check there).

Do you know if there's any other checks to prevent pvp happening in rookgard? (In dawnport I assume that people would have vocations there, but there is any form to stay in dawnport as no-vocation?)

dudantas:
@andersonfaaria, It is only possible to be in dawnport in combat zones with a vocation.

ghost:
@andersonfaaria as @dudantas said on Dawnport it's only possible to be in combat zones with a vocation.
I honestly think it isn't suitable to have map settings on source files (as wether Rookgaard is or isn't a non-PVP zone) but to allow backward compatibility (and in answer to the feedback received) I'll add a Lua script to disable combat between characters with no vocation on my PR (as well as on OTLand's TFS) so this PR shouldn't be approved yet.

majestyotbr:
PR closed due to inactivity. If you want, you can reopen to finish.

PR #1607, opened 2020-07-31
new feat: Map click teleport

ghost:
Added teleport by CTRL+Shift+click on minimap or on Cyclopedia map.

Also there's a new custom flag on groups.xml to define which groups can use this feature so server owners don't need to edit source files to change that.

About why implementing custom player flags instead of a new player flag: this is meant to keep compatibility with TFS, so even if new player flags get added on it we will not be forced to add them with different exponential values than on TFS and we will not mix original flags with flags added for new features on OTServBR-Global like this one.

### **IMPORTANT NOTE:**
All credits to this new feature should go to @jo3bingham and [his post](https://otland.net/threads/11-50-an-easier-way-to-teleport.268531/) on OTLand. I merely added his code to OTServBR-Global source and created a custom flag for easier configuration for server owners.

### DIFF
--- a/data/XML/groups.xml
+++ b/data/XML/groups.xml
@@ -56,6 +56,9 @@
 			<flag cannotbemuted="1" />
 			<flag isalwayspremium="1" />
 		</flags>
+		<customflags>
+			<customflag canmapclickteleport="0" />
+		</customflags>
 	</group>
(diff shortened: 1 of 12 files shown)


Geekbuddy:
Nice PR @kairion ! I gonna test it.

majestyotbr:
Tested, working. Thanks for the contribution.

majestyotbr (approved):

PR #1617, opened 2020-08-04
Fix: magic gold converter + gold pouch

### LINKED ISSUE #1448: Gold pouch visual glitch when using the magic gold converter.
**Describe the bug**
After the commits of July 6th, whenever you loot more then 100 gold and the magic converter is turned on, after converting the gold stays visually on your pouch and will only disappear when you loot more gold.

**To Reproduce**
1. Assign the gold to be looted to your gold pouch and have it opened.
2. Loot 50 gold.
3. Loot 100+ gold.
4. Check the gold pouch window, the 100 stack will be there and it doesn't even exist anymore. Sometimes the bug is even worse and it creates two stacks of the same amount of gold, even thou they both don't really exist.
5. Close the pouch and reopen it, the gold will not be there anymore and the converted gold will be there.

6. Also, the converted gold will be thrown in your default backpack, not your gold pouch, which I think it's also a bug. If that could be fixed as well it would be great.

**Expected behavior**
The conversion should happen visually as well, no matter how many gold I loot.

**Screenshots**
This is after I looted the 100+ gold and the magic converter converted the gold.
gold_pouch_before_reopen

This is after I reopened the gold pouch, you can see that the gold got properly converted.
As said in step number 6, the converted gold was thrown in my default backpack. Even thou all gold should be sent to my gold pouch.
gold_pouch_after_reopen

**Additional context**
As I said in the beginning of the issue, this started happening after the commits of July 6th, so it might be worth investigating.

marcosvf132:
- [x] Visual bug (FIXED)
Resolves #1448
 On the past `player.cpp` changes was implemented a script related to ITEM_GOLD_POUCH, the changes was making the gold pouch generate some visual bugs on moving coin to the pouch. With this commit, now the gold pouch works like a regular backpack, but maintain the restriction to just move coins to it.

### DIFF
--- a/data/actions/scripts/others/magic_gold_converter.lua
+++ b/data/actions/scripts/others/magic_gold_converter.lua
@@ -13,23 +13,20 @@ local function start_converter(pid, itemid)
 				if charges_n >= 1 then
 					local quickLootBackpacks = player:getQuickLootBackpacks()
 					if player:getItemCount(ITEM_GOLD_COIN) >= 100 then
-						local goldcount = math.floor(player:getItemCount(ITEM_GOLD_COIN) / 100)
-						player:removeItem(ITEM_GOLD_COIN, goldcount * 100)
-						local platitem = player:addItem(ITEM_PLATINUM_COIN, goldcount)
+						player:removeItem(ITEM_GOLD_COIN, 100)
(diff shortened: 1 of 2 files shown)


majestyotbr:

majestyotbr (approved):

PR #1623, opened 2020-08-05
Update golden outfit memorial

cys9211:
libs storages

ShortcutHubDoor = 52000,

GoldenOutfit = 51015, - memorial work on this storage!

### DIFF
--- a/data/actions/scripts/others/golden_outfit_memorial.lua
+++ b/data/actions/scripts/others/golden_outfit_memorial.lua
@@ -1,4 +1,4 @@
-local storage = 52000
+local storage = 51015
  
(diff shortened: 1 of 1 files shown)


data/actions/scripts/others/golden_outfit_memorial.lua
```diff
@@ -1,4 +1,4 @@
-local storage = 52000
+local storage = 51015
```
marcosvf132: Its better to work with global `Storage.OutfitQuest.GoldenOutfit` instead numbers. To prevent further problems just like that.

majestyotbr: I agree.

majestyotbr (approved):

PR #1624, opened 2020-08-05
Feat: Cobra flask caldron

marcosvf132:
- After using `flask with snake poison` on the `Large Cauldron`, the Cobra monsters will get a debuff on their health for the next 30 minutes.
https://tibia.fandom.com/wiki/Flask_with_Snake_Poison

- To properly work, a map change need to be done too **at position {x = 33399, y = 32662, z = 5}:** 

1. Large caldron itemID: from **ID** 36119 to **ID** 36122. (4 raw parts)
2. Empty flask itemID: 36132.

caldron
Works better when testing with a regular character.
The flask is a daily respawn, so it should be on the map.

### DIFF
--- a/data/actions/actions.xml
+++ b/data/actions/actions.xml
@@ -65,6 +65,7 @@
 	<action itemid="23422" script="oramond/elevator.lua" />
 	<action itemid="23429" script="oramond/elevator.lua" />
 	<!-- others -->
+	<action fromid="36131" toid="36132" script="others/cobra_flask.lua" />	
 	<action itemid="6558" script="others/concentrated_demonic_blood.lua" />
 	<action fromid="11258" toid="11262" script="others/blessing_charms.lua" />
 	<action fromid="28036" toid="28037" script="others/blessing_charms.lua" />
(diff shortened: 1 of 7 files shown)


attack787:
I tested, it works sometimes and with some creatures. For example, I put the poison in the boiler, the message from the buff appears. All right, After I kill the snake creatures (vizier, scout and assassin) when they are born, some are born with 70% of life, others with full life. It appears to be random and does not affect all creatures in the respawn.

marcosvf132:
> I tested, it works sometimes and with some creatures. For example, I put the poison in the boiler, the message from the buff appears. All right, After I kill the snake creatures (vizier, scout and assassin) when they are born, some are born with 70% of life, others with full life. It appears to be random and does not affect all creatures in the respawn.

Are you testing with GOD character? This only happens to me when i'm on a GOD char. The onThink function usually only triggers when there is a oponnent (regular char) near, its good to avoid system overload

attack787:
> > I tested, it works sometimes and with some creatures. For example, I put the poison in the boiler, the message from the buff appears. All right, After I kill the snake creatures (vizier, scout and assassin) when they are born, some are born with 70% of life, others with full life. It appears to be random and does not affect all creatures in the respawn.
> 
> Are you testing with GOD character? This only happens to me when i'm on a GOD char. The onThink function usually only triggers when there is a oponnent (regular char) near, its good to avoid system overload

Sorry, I did test with god. It works as described. Thanks for sharing!

MajestyOTBR:
@dudantas PR waiting for your review.

majestyotbr (approved):

PR #1626, opened 2020-08-06
Create golem.lua

cys9211:


cys9211:
Combine #1626  with #1627

oualid6496:
Does it work like that on real to?

cys9211:
> Does it work like that on real to?

In the real is a Reward from Shadows of Yalahar Quest.

Unfortunately, this mission is not and will not be in the project because it is unnecessary

oualid6496:
> > Does it work like that on real to?
> 
> In the real is a Reward from Shadows of Yalahar Quest.
> 
> Unfortunately, this mission is not and will not be in the project because it is unnecessary

Says who? I thougt it was supposed to look as much as rl as possible.

cys9211:
> > > Does it work like that on real to?
> > 
> > In the real is a Reward from Shadows of Yalahar Quest.
> > 
> > Unfortunately, this mission is not and will not be in the project because it is unnecessary
> 
> Says who? I thougt it was supposed to look as much as rl as possible.



I just added the script, the rest was already on the map

oualid6496:
I think it's better to remove it from the map instead of making the script for it

majestyotbr:
#726 Quests that need to be developed or revised.

Unfortunately, the script is not good. If you want to contribute, follow the pattern of the scripts of the project and do in one pull request.

PR #1627, opened 2020-08-06
Adding script Crystal Sword change to golem wrench

cys9211:


majestyotbr:
#726 Quests that need to be developed or revised.

If you want to contribute, follow the pattern of the scripts of the project and do in one pull request.

PR #1636, opened 2020-08-09
Feat: Supply Stash

marcosvf132:
Some parts I got from **vinioliveirasilva** and **SaiyansKings** script, **special thanks to them!**
- **ATM this is working with otservbr client 12.40+, on custom clients you MAYBE can have bugs and/or crashes, if this happens to you, tell me! maybe some changes might need to be done to support custom/old clients.**
- [x] Stow specific item by menu.
- [x] Stow all player items of same type.
- [x] Stow containers.
- [x] Drag and drop to stow.
- [x] Market integration.
- [x] Hireling NPC integration.
- [ ] Quickloot retrieve backpack. (Need quickloot cpp to be implemented without further conflicts)

- On default, all stackable items can be stowed, but there is a black list items that cannot be stowed. Add this line on items.xml to include a stackable item on the blacklist:
### If you find a item that sould not be stowed into the supply stash, please leave a comment here.
``` xml
<attribute key="notstowable" value="1" />
```
- Need to be tested all **stash functions** and **market**, since there is few changes on market to implement the market-stash integration.


attack787:
Something bizarre happened using this system. When I used the supply stash for the first time, I put my bp on supply stash and I opened the supply stash to check what items that were stored inside and the client crashed (cliente crash for two times). I went to the db to look at the items that were placed in the stash and only had the sinister book (id: 32705). I deleted this item in db and the supply back to work again. I tried to put this item in the supply stash again and it didn't happen again. I logged another char into another account to make test and it didn't happen again. As I said at the beginning, it was bizarre and I'm here reporting the fact. I will continue to make more tests to find any bugs. Gratz for the work and thanks for the contribution!

marcosvf132:
> Something bizarre happened using this system. When I used the supply stash for the first time, I put my bp on supply stash and I opened the supply stash to ...

The script store the item clientID, thats why the item on DB was a weird one. Which item did you try to stow? it was stackable? you tried to stow it again and it didn't crash? please let me know if you remember the item so i can run more tests too, ty! Usualy the client crash when the id on DB is a non-stackable or 'itemid' instead 'clientid', but the script only save the item by 'clientid', idk why your client crashed, if this happen again to you, try to remember the item name and which id is on your DB, so i can try to stow it too.

attack787:
> > Something bizarre happened using this system. When I used the supply stash for the first time, I put my bp on supply stash and I opened the supply stash to ...
> 
> The script store the item clientID, thats why the item on DB was a weird one. Which item did you try to stow? it was stackable? you tried to stow it again and it didn't crash? please let me know if you remember the item so i can run more tests too, ty! Usualy the client crash when the id on DB is a non-stackable or 'itemid' instead 'clientid', but the script only save the item by 'clientid', idk why your client crashed, if this happen again to you, try to remember the item name and what id is on your DB, so i can try to stow it too.


The item names "sinister book - id: 32705" it is not stackable and is not part of the supply stash. What I found strange was that in the first use of the supply stash the system accepted this item and after deleting it in the db "player_stash" and trying to put it back in the supply stash, it was no longer added.

marcosvf132:
But you did try to stow sinister book? or you just saw the id on DB? because as i said, the script convert the regular itemid to clientid, Eg: avalanche rune **itemid** is 2274 and **clientid** is 3161. The script store the clientid on the db.

attack787:
> But you did try to stow sinister book? or you just saw the id on DB? because as i said, the script convert the regular itemid to clientid, Eg: avalanche rune **itemid** is 2274 and **clientid** is 3161. The script store the clientid on the db.


Now I understand what you sayd. I opened the item editor and looked for the item. So  idclient: 32705; itemid: 37540; (name: pair of old bracer). This items makes client crash when you try open the stash. I'm using cliente 12.20 with sprites 12.40, maybe its that is the reason for the crashes. If you have client 12.40, do test and see if the same problem occurs.

marcosvf132:
image
Your client is causing the crash, on otservbr 12.40 client this item can be stowed with no problem at all.

attack787:
> image
> Your client is causing the crash, on otservbr 12.40 client this item can be stowed with no problem at all.

ok, its good. It would be nice to inform that the system works only on 12.40 clients.

marcosvf132:
When I have more time i will try to make changes to support all old/custom clients.

src/const.h
```diff
@@ -557,9 +557,6 @@ enum item_t : uint16_t {
 	ITEM_CRYSTAL_COIN = 2160,
 	ITEM_STORE_COIN = 24774,
 
-	ITEM_DIAMOND_ARROW = 29057,
-	ITEM_SPECTRAL_BOLT = 29058,
```
gpedro: why you removed that? is crashing the compilation

marcosvf132: I removed the itemID check from player.cpp, now is the new function isStowable() that is checking if the item can be stowed or not

marcosvf132: I added it on the last change, but using isStowable i think is better to control the items without the need of compile again

marcosvf132: Check if you have the new player.cpp change, maybe thats why its crashing for you

PR #1637, opened 2020-08-09
Add ghost backpack to imbuement and movements

cys9211:
Adding ghost backpack (imbu no working)


majestyotbr:
It's not missing to add the id 37455 of ghost backpack in imbuements.lua?
https://github.com/opentibiabr/otservbr-global/blob/develop/data/lib/core/imbuements.lua#L16

PR #1644, opened 2020-08-11
Fix: startup functions.lua

marcosvf132:


data/startup/others/functions.lua
```diff
@@ -9,8 +9,8 @@ function loadLuaMapAction(tablename)
 			if tile then
 				-- Checks that you have no items created
 				if tile:getItemCountById(value.itemId) == 0 then
-					print(">> Wrong item id found")
-					print(string.format("> Action id: %d, item id: %d", key, value.itemId))
```
dudantas, line 13: What is the reason for removing the log?
In this case, the idea is that if the item set in the itemId variable does not exist on the map, it returns saying that the id is wrong.

data/startup/others/functions.lua
```diff
@@ -49,8 +49,8 @@ function loadLuaMapUnique(tablename)
 		if tile then
 			-- Checks that you have no items created
 			if tile:getItemCountById(value.itemId) == 0 then
-				print(">> Wrong item id found")
-				print(string.format("> Unique id: %d, item id: %d", key, value.itemId))
```
dudantas, line 53: Same here

PR #1661, opened 2020-08-14
Cobra flask respawn

marcosvf132:
On the previous change, the empty flask was added only on the map, but It suppouse to be on map and respawn every 8 hours. This PR add the respawn if the item is not on the position.
image

### DIFF
--- a/data/scripts/globalevents/cobra_flask.lua
+++ b/data/scripts/globalevents/cobra_flask.lua
@@ -1,0 +1,14 @@
+local position = Position(33395, 32666, 5)
+
+local cobraEmptyFlask = GlobalEvent("Cobraflask")
+
+function cobraEmptyFlask.onThink(interval)
+local flask = Tile(positions):getItemById(36132)
+if not(flask) then
+	Game.createItem(36132, 1, position)
+end
+	return true
+end
+
+cobraEmptyFlask:interval(1000*60*60*8) -- 8 hours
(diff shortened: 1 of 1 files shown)


data/scripts/globalevents/cobra_flask.lua
```diff
@@ -0,0 +1,14 @@
+local position = Position(33395, 32666, 5)
+
+local cobraEmptyFlask = GlobalEvent("Cobraflask")
+
+function cobraEmptyFlask.onThink(interval)
+local flask = Tile(positions):getItemById(36132)
```
dudantas: The correct is "position"`and not "positions" 
But, you can remove the variable and put:
````local flask = Tile({x = 33395, y = 32666, z = 5}):getItemById(36132)````

dudantas (changes requested):

data/scripts/globalevents/cobra_flask.lua
marcosvf132: ops

attack787:
Don't work for me. I removed the item of map editor and used this script. Nothing happens

marcosvf132:
Dont need to remove the item from your map, this script only respawn the item every 8 hours, so if you remove it, it will spawn again 8 hours later.

attack787:
I reduced the time to test and this is working, thanks.

majestyotbr (approved):

PR #1670, opened 2020-08-17
Svargrond arena fixes

### LINKED ISSUE #164: svarground arena bug
you go with the npc you say greenhorn and he doesn't let pass

### LINKED ISSUE #1610: svarground arena bug
you go with the npc you say greenhorn and he doesn't let pass

dudantas:
Resolves #1610

Fixed bug that did not reset the door storage (from Halvar) when leaving the arena
Fixed bug where it was possible to use the trophy and write on the item
Fixed bug of wrong position of the last door, so it didn't set the action
Fixed bug that did not allow entering the doors of the Halvar npc


attack787:
Tested and It's working. Thanks for fix this bug

PR #1802, opened 2020-10-02
Fix market and npc trade on protocol 12.51

FakeShinoda:
- Fix market debug
- Fix npc trade debug

_Credits to: @SaiyansKing for these functions._

Enjoy.. AND TEST IT !


majestyotbr:
Tested, working. Thanks for the fix!

majestyotbr (approved):

PR #1830, opened 2020-10-06
Add soul weapons to movements

yatozin:

PR #1832, opened 2020-10-06
Update diamond_arrow.lua

cys9211:
Diamond Arrow is use a BOW!


marcosvf132:
This fix is already on PR #1830. But thank you for your contribution!!

PR #1833, opened 2020-10-06
Fix: 'move up' stackable item

### LINKED ISSUE #204: Bug on "Move up"
When you try to move up something the mensage "sorry not is possible " appear and the function doenst work

a video to explain it better:

https://www.youtube.com/watch?v=VmkDdtHeeFM

### LINKED ISSUE #666: Bug on "Move up"
When you try to move up something the mensage "sorry not is possible " appear and the function doenst work

a video to explain it better:

https://www.youtube.com/watch?v=VmkDdtHeeFM

marcosvf132:
Resolves #666 :O
 For some reason the client returns count = 255 even on stackable items when using the 'move up' menu function, so to fix this bug we need to recalculate the item count. As this 'move up' option is used to move all items on that specific slot to the item parent container, there is no problem using item:getCount() to move all count.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1393,7 +1393,8 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 	}
 
 	// 'Move up' stackable items fix
-	if (item->isStackable() && count == 255) {
+	//  Cip's client never sends the count of stackables when using "Move up" menu option
+	if (item->isStackable() && count == 255 && fromCylinder->getParent() == toCylinder) {
 		count = item->getItemCount();
 	}
 
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1392,6 +1392,11 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 		return RETURNVALUE_NOERROR; //silently ignore move
 	}
 
+	// 'Move up' stackable items fix
+	if (item->isStackable() && count == 255) {
+		count = item->getItemCount();
+	}
+
 	//check if we can add this item
 	ReturnValue ret = toCylinder->queryAdd(index, *item, count, flags, actor);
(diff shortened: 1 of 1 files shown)


majestyotbr:
@debone What do you think?

debone:
I think the fix is correct with one modification. 

After trying to understand everything I arrived at the same conclusion. The `Game::internalMoveItem` is trying to move 255 items, which is not possible, thus the error. Given that the "Move up" arrives to the server as a simple move, we cannot simply remove the checks later on that would also fix this, but break other behaviours. To be real safe, I thought that we could add another check to the if `fromCylinder->getParent() == toCylinder`, so the chances that only the correct "Move up" would get by is higher in such generic function. In the end, it actually caught an issue with depots and trying to Move up items to the main box, creating the correct "Sorry, not possible" message. Without this extra check, it's still not possible to move up anything in depots, but there is an attempt to move the item and it only falls back to the container again, making it look like nothing happened.

src/game.cpp
```diff
@@ -1392,6 +1392,11 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 		return RETURNVALUE_NOERROR; //silently ignore move
 	}
 
+	// 'Move up' stackable items fix
+	if (item->isStackable() && count == 255) {
```
debone:
```suggestion
	if (item->isStackable() && count == 255 && fromCylinder->getParent() == toCylinder) {
```

src/game.cpp
```diff
@@ -1392,6 +1392,11 @@ ReturnValue Game::internalMoveItem(Cylinder* fromCylinder, Cylinder* toCylinder,
 		return RETURNVALUE_NOERROR; //silently ignore move
 	}
 
+	// 'Move up' stackable items fix
```
debone, line 1395:
```suggestion
	// 'Move up' stackable items fix
	//  Cip's client never sends the count of stackables when using "Move up" menu option
```

majestyotbr (approved):

PR #1838, opened 2020-10-07
Fixing stackables not being able to be added to full containers

### LINKED ISSUE #200: Stack bug
You can't take the stackable item from the ground if your backpack got all slots used and you have this stackable item.
Example:
1 gmp laying on the ground
Player Has 19 Items and 50 gmp in the backpack.
Player tries to pick up 1 gmp and put it into 50 gmp in the backpack.
Player gets an error saying he cannot do it

### LINKED ISSUE #1748: Stack bug
You can't take the stackable item from the ground if your backpack got all slots used and you have this stackable item.
Example:
1 gmp laying on the ground
Player Has 19 Items and 50 gmp in the backpack.
Player tries to pick up 1 gmp and put it into 50 gmp in the backpack.
Player gets an error saying he cannot do it

debone:
Fixes #1748, **but I have no idea if it is the right approach**. If it's ok, I need to remove the comments from the code.

I removed a check that happened when adding stackables to a full container. The check was calculating if the player could carry the full remaining stackable amount, not what the amount the player wanted to add. 

For example, consider that the player have a container full of ropes and 1 GMP. When the player tries to add another GMP to the container, `queryAdd` tries to see if the player can carry the remaining 100-1 = 99 GMPs, even though the player just wants to add 1. In #1748 what happens is that the player doesn't have the capacity to carry the full amount of the remaining amount (99 GMPs in the example), as `queryAdd` refuses to let the player add it to the container, even though the player would have capacity for just 1 more GMP. 

I think `queryAdd` is unecessary here and I cannot tell if it breaks anything else if it's not present. I tested and if you try to add a resonable amount of items to a full container with the stack, it works. If you try to add too many (no capacity), it fails.

### DIFF
--- a/src/container.cpp
+++ b/src/container.cpp
@@ -412,12 +412,7 @@ ReturnValue Container::queryMaxCount(int32_t index, const Thing& thing, uint32_t
 		} else {
 			const Item* destItem = getItemByIndex(index);
 			if (item->equals(destItem) && destItem->getItemCount() < 100) {
-				uint32_t remainder = 100 - destItem->getItemCount();
-				n = remainder;
-				// Here is "Can we add 98 (ultra heavy) items to player?"
-				// When we could be asking "Can we add the amount players want?"
-				// if (queryAdd(index, *item, remainder, flags) == RETURNVALUE_NOERROR) {
-				// }
+				n = 100 - destItem->getItemCount();
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


MajestyOTBR:
Maybe the issue was fixed before, because i didn't confirmed here.

MajestyOTBR:
Tested again with updated repository and compatible executable, can't reproduce the issue.

debone:
Hum, I just checked on latest repository and it's still there. It's a weird bug, please reconsider :). I suspect you run the test on a cap unlimited character, as I also have initially tested with a "God" player and the bug doesn't manifest as there's a lot of cap.

Below another attempt to explain what's going on
CleanShot 2020-10-10 at 23 22 53@2x

majestyotbr:
I tested with a normal character and enough cap, which was not the case.
Thanks for clarify the problem, i tested again and with the fix now its working as intended.

majestyotbr (approved):

majestyotbr:
Now you can remove the comments from the code.

PR #1842, opened 2020-10-08
Fix #923

### LINKED ISSUE #923: Paralyze Rune Bug
The Paralyze Rune is not working as intended, in RL Tibia you can heal paralyzed effect while moving, but in OTServ you can't.

How to Reproduce:
- Have an ED to paralyze you
- Try to heal paralyzed condition **while moving**

dracoalv:
- Fix paralyze rune behaviour.

### DIFF
--- a/data/spells/scripts/support/paralyze rune.lua
+++ b/data/spells/scripts/support/paralyze rune.lua
@@ -1,16 +1,30 @@
-local combat = Combat()
-combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_MAGIC_RED)
-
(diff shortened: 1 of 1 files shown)


debone:
@dracoalv do you plan to work more on this pull request? It seems that the behaviour, described in this conversation https://discordapp.com/channels/528117503952551936/576786020067049472/765961247630950401, is that the paralyze rune overrides any kind of velocity condition and fixes the velocity on 40, which doesn't seem to be covered on your contribution.

debone:
I will take that as a no :) Feel free to contribute with the correct fix :)

PR #1845, opened 2020-10-09
Fix: pzLocked on field rune (PVP)

### LINKED ISSUE #197: PK on PZ
If a player uses a field rune outside of a protection zone, proceeds to enter the pz and someone else steps on it, the player will get a skull and be able to move around inside of the pz area.

This is not a big deal until other players start dying on noob chars, resulting in a red and possibly black skull, while inside of a protected zone.

1- Player A uses energy bomb outside of temple
2- Player A enters temple (pz area)
3- Player B should try to step in and out of the energy field until hit by it
4- Player A receives white skull inside of the temple and can still move around

Possible fixes:
- While player is inside of a pz, or has the pvp hand open, the field should not give him a pk, nor hit other players.

- Player should get a pz lock when using a field rune (used to be like this but isn't practical anymore, with the pvp changes)

### LINKED ISSUE #946: PK on PZ
If a player uses a field rune outside of a protection zone, proceeds to enter the pz and someone else steps on it, the player will get a skull and be able to move around inside of the pz area.

This is not a big deal until other players start dying on noob chars, resulting in a red and possibly black skull, while inside of a protected zone.

1- Player A uses energy bomb outside of temple
2- Player A enters temple (pz area)
3- Player B should try to step in and out of the energy field until hit by it
4- Player A receives white skull inside of the temple and can still move around

Possible fixes:
- While player is inside of a pz, or has the pvp hand open, the field should not give him a pk, nor hit other players.

- Player should get a pz lock when using a field rune (used to be like this but isn't practical anymore, with the pvp changes)

marcosvf132:
Resolves #946
 Uppon using field runes and with the world type set to 'pvp' (retro-pvp), the user should get pzlocked. Before this PR this feat was missing.

Note: By default, if a player use a field rune (ex: fire bomb rune) on another player, then the first player will get white skull, this PR only affects when the first player uses the field rune on anywhere except on another player.

### DIFF
--- a/data/spells/spells.xml
+++ b/data/spells/spells.xml
@@ -455,18 +455,18 @@
 	</instant>
 
 	<!-- Attack Runes -->
-	<rune group="attack" spellid="26" name="Poison Field" id="2285" allowfaruse="1" charges="3" lvl="14" maglv="0" exhaustion="2000" groupcooldown="2000" blocktype="solid" script="attack/poison field.lua" />
-	<rune group="attack" spellid="91" name="Poison Bomb" id="2286" allowfaruse="1" charges="2" lvl="25" maglv="4" exhaustion="2000" groupcooldown="2000" blocktype="solid" script="attack/poison bomb.lua" />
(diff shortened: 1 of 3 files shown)


majestyotbr (approved):

adrianomengue:
@marcosvf132 could you take a lookt at #2511 ? Its related with your PR.

PR #1855, opened 2020-10-10
Fix waiting list

heberholanda:
**Credits: TFS**


majestyotbr:
The problem still happens.

PR #1872, opened 2020-10-12
Removing cpplint rule for tab vs spaces

debone:
I'm removing given plenty of PRs get through with it plainly failing, and it could be nice to take it more serious.

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -39,6 +39,8 @@ jobs:
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         reporter: github-pr-check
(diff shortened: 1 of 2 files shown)


Costallat:
I think that it's better to start fixing what cpplint is complaining than start to disable the rules

debone:
I wholeheartedly agree with you @Costallat. I followed the practical way and wanted to clean up the noise to be able to see something on the PRs, starting slower. Can you take a decision here and maybe we will not accept PRs with cpplint failing then?

Costallat:
To keep PRs 'clean' and how we know that people will not fix those things it's better to remove it to help us.

Costallat (approved):

Issue #1880, opened 2020-10-14
Paralyze condition is not 100%


morenao97:
On Global Tibia, when you get a Paralyze condition, your speed is fixed in a value, and independent such u have a equipped item, or not, your speed will be that!
This video show that happens when i equip a boh, when i'm paralyzed, my speed not change.
https://imgur.com/XH6qDE2

PR #1881, opened 2020-10-14
Fix: Paralyze Rune Formula

morenao97:
The Paralyze rune, on Tibia global, changes ur speed to 40, not 22.

A video showing this: https://imgur.com/XH6qDE2

### DIFF
--- a/data/spells/scripts/support/paralyze rune.lua
+++ b/data/spells/scripts/support/paralyze rune.lua
@@ -3,7 +3,7 @@ combat:setParameter(COMBAT_PARAM_EFFECT, CONST_ME_MAGIC_RED)
 
 local condition = Condition(CONDITION_PARALYZE)
 condition:setParameter(CONDITION_PARAM_TICKS, 20000)
-condition:setFormula(-1, 45, -1, 45)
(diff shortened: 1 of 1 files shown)


debone:
Thanks for the submission @morenao97! It's great to have more contributors in the project :)

As I remember, the paralyze rune has a overwrite effect, keeping the speed at 40 no matter which speed conditions are altered (like wearing a BOH or a Time ring, right?). I don't see that happening in the way you implemented the change with the formula. Are you planning to work more on this fix?

morenao97:
> Thanks for the submission @morenao97! It's great to have more contributors in the project :)
> 
> As I remember, the paralyze rune has a overwrite effect, keeping the speed at 40 no matter which speed conditions are altered (like wearing a BOH or a Time ring, right?). I don't see that happening in the way you implemented the change with the formula. Are you planning to work more on this fix?

No! but i already have created a issue ([Link](https://github.com/opentibiabr/otservbr-global/issues/1880)), all of the paralyzes(from players or monsters) on global have this 'effect' of overwrite, and isn't possible to add on our project without change something on sources

debone:
Great. I will close this PR then.

PR #1882, opened 2020-10-15
Format source code using Google Style

Costallat:


src/baseevents.cpp
```diff
@@ -26,185 +26,181 @@
...
-	}
-	return true;
+bool Event::checkScript(const std::string& basePath,
+                        const std::string& scriptsName,
+                        const std::string& scriptFile) const {
+  LuaScriptInterface* testInterface = g_luaEnvironment.getTestInterface();
+  testInterface->reInitState();
+
+  if (testInterface->loadFile(
+          std::string(basePath + "lib/" + scriptsName + ".lua")) == -1) {
```
debone, line 111: I understand that this if has a line break just to keep it under the line lenght limit, right?

debone:
Is there something more than moving the source from 1 tab to 2 spaces?

PR #1888, opened 2020-10-16
Fix #1250 - Summons with targetdistance running away from target

### LINKED ISSUE #1250: Summons with targetdistance running away from target
Summons that are supposed to attack from distance (targetdistance=">0") are running away "infinitely" (until target is lost for their master), if attacked monster target them.
[Vid showing that problem.](https://gfycat.com/importantfastdesertpupfish)

To reproduce:
1. Summon creature that attack from distance (emberwing, minotaur archer etc)
2. Attack hostile monster
3. Let that monster target your summon instead of you
4. Your summon is running away like crazy

Expected behavior:
Your summon is walking one tile every 2 sec. [Video showing that](https://youtu.be/68AVCNMZ8Bs?t=490) - skullfrost is slowly walking away instead of running.

I tried to fix that myself for a while, but failed as my attempt didnt work.

Code to look at:
https://github.com/opentibiabr/otservbr-global/blob/9965cd24f4ce5b9851e6e95dfc2d7e1cbde3b574/src/creature.cpp#L895
https://github.com/opentibiabr/otservbr-global/blob/9965cd24f4ce5b9851e6e95dfc2d7e1cbde3b574/src/monster.cpp#L1361
https://github.com/opentibiabr/otservbr-global/blob/ea1d34d4a6b8a88a4eec6712c082a430eb4a4cde/src/monster.cpp#L805

heberholanda:


yatozin:
when using this with the follow fix the summon just stands still when you target a monster, and teleports if you go out of range but then stand still again when target.

gerotib:
Fix like this is not solving the source of problem, because they are supposed to attack from distance

PR #1895, opened 2020-10-18
Accessory conditions are not being applied

debone:
For example scorpion melee attacks always get you poisoned

CleanShot 2020-10-18 at 19 56 55@2x

### DIFF
--- a/src/monsters.cpp
+++ b/src/monsters.cpp
@@ -597,24 +597,6 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
 				sb.maxCombatValue = -Weapons::getMaxMeleeDamage(spell->skill, spell->attack);
 			}
 
-			ConditionType_t conditionType = CONDITION_NONE;
-			int32_t minDamage = 0;
-			int32_t maxDamage = 0;
-			uint32_t tickInterval = 2000;
-
-			if (spell->conditionType != CONDITION_NONE) {
-				conditionType = spell->conditionType;
-
-				minDamage = spell->conditionMinDamage;
-				maxDamage = minDamage;
-				if (spell->tickInterval != 0) {
-					tickInterval = spell->tickInterval;
-				}
-
-				Condition* condition = getDamageCondition(conditionType, maxDamage, minDamage, spell->conditionStartDamage, tickInterval);
-				combat->addCondition(condition);
-			}
-
 			sb.range = 1;
 			combat->setParam(COMBAT_PARAM_TYPE, COMBAT_PHYSICALDAMAGE);
 			combat->setParam(COMBAT_PARAM_BLOCKARMOR, 1);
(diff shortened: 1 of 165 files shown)


src/monsters.cpp
```diff
@@ -736,6 +695,36 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
 			combat->setParam(COMBAT_PARAM_EFFECT, spell->effect);
 		}
 
+        // A condition is a condition is a condition
```
lgrossi: This comment is weird, haha
Also, indentation here looks a bit weird

src/monsters.cpp
```diff
@@ -736,6 +695,36 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
...
+        // If a spell has a condition, it always applies, no matter the origin
+        if (spell->conditionType != CONDITION_NONE) {
+            int32_t minDamage = std::abs(spell->conditionMinDamage);
+			int32_t maxDamage = std::abs(spell->conditionMaxDamage);
+			int32_t startDamage = 0;
+			uint32_t tickInterval = 2000;
+
+			if (spell->tickInterval != 0) {
```
lgrossi: Why not simplifying that?
Either
```
  if (spell->tickInterval > 0) {
    tickInterval = spell->tickInterval;
  }
```
Or
```
tickInterval = spell->tickInterval > 0 ? spell->tickInterval : 2000; //here you can also remove the initial declaration
```

src/monsters.cpp
```diff
@@ -736,6 +695,36 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
...
+				int32_t value = spell->tickInterval;
+				if (value > 0) {
+					tickInterval = value;
+				}
+			}
+
+			if (spell->conditionStartDamage != 0) {
+				int32_t value = std::abs(spell->conditionStartDamage);
```
lgrossi: Also here
```
int32_t startDamage = std::abs(spell->conditionStartDamage);
if (startDamage > minDamage) {
  startDamage = 0;
}
```

src/monsters.cpp
```diff
@@ -736,6 +695,36 @@ bool Monsters::deserializeSpell(MonsterSpell* spell, spellBlock_t& sb, const std
...
+				if (value <= minDamage) {
+					startDamage = value;
+				}
+			}
+
+            if (maxDamage == 0) {
+                maxDamage = minDamage;
```
lgrossi: Id use ternary here also, but this is fine.

lgrossi (commented):
Nice spot, I think you could improve the cpp lart since you're refactoring, though. Code was made in a confusing way before, imho, simplifying makes it a bit more understandable. A lot of lines not always mean better clarity.

src/monsters.cpp
debone: HAHA, this was me talking to the past given they were ignoring conditions just because. If there's a condition, it's a condition!

I see the indentation, will fix.

PR #1903, opened 2020-10-22
Adding increased logs for lua call stack overflows

debone:


lgrossi:
Would you be so kind to move those couts to spdlog? It would improve a lot the logging.  Nice job though!

heberholanda (approved):

PR #1913, opened 2020-10-26
Fixing respawn type on monsters revsys

debone:
- Breaking change for better respawnType API on lua scripts
- Removed old respawnType flag on monsters
- Deprecated SpawnType_t enum
- Backward compability with XML monsters
- Adding Faun lua sample

### DIFF
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -108,8 +108,8 @@ registerMonsterType.flags = function(mtype, mask)
 		if mask.flags.passive then
 			mtype:isPassive(mask.flags.passive)
 		end
-		if mask.flags.respawntype then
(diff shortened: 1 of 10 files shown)


dex-89:
Dark faun spawn only in night but there are in cave in day too... Remember this when writing this script.

I don't saw option: underground... Sorry :)

debone:
> Dark faun spawn only in night but there are in cave in day too... Remember this when writing this script.
> 
> I don't saw option: underground... Sorry :)

Don't worry! :) I was wondering actually what was the logic of the flag given I'm not playing much right now. 

If I understood you correct then, if I mark some creature `underground`, it will spawn regardless if it's day or night. And if this creature is also nocturn, then it will spawn only during the night **or** the whole day in caves.

PR #1923, opened 2020-11-01
Fix #1905 - MoveEvent().onRemoveItem doesn't work

### LINKED ISSUE #1905: MoveEvent().onRemoveItem doesn't work
**Describe the bug**

When moving the parchment aid(104) nothing happens. 
Apparently the method isn't executed.

..data/scripts/quests/parchment_room/parchment.lua

**To Reproduce**
Move to pos (33064, 31624, 15) Quest Key 6010.

**Expected behavior**
When moving the parchment it should summon the other demons.

**Screenshots**
image

heberholanda:
Action AID changed


data/lib/core/functions.lua
```diff
@@ -9,6 +9,10 @@ function getDistanceBetween(firstPosition, secondPosition)
 	return posDif
 end
 
+function doComparePositions(position, positionEx)
```
debone: Maybe reuse https://github.com/opentibiabr/otservbr-global/blob/develop/data/lib/core/position.lua ?

data/events/scripts/player.lua
```diff
@@ -314,6 +314,37 @@ end
 
 function Player:onMoveItem(item, count, fromPosition, toPosition, fromCylinder, toCylinder)
```
debone, line 315: While I agree to use a different API to check the parchment, I don't think moving it out of its own file to the player.lua is a good move. We cannot keep such specific functionality for a quest mixed with the player stuff.

debone (changes requested):

data/events/scripts/player.lua
heberholanda: I don't like it either, do you have any suggestions?

data/lib/core/functions.lua
heberholanda: Yes, I change now.

dudantas:
In reality, onRemoveItem bugged this PR: https://github.com/opentibiabr/otservbr-global/pull/1779/files

The solution is only for onAddItem. But, thanks.

PR #1927, opened 2020-11-02
Items Max Limit

heberholanda:


dex-89:
Why exactly 80000?

heberholanda:
Apparently items.otb has a limit of 40000 items, this change is to support more items from the new updates. We have already reached that limit.

debone:
Now that I see this PR, I think we need to discuss more details about it and how the fix will look like.

debone:
Let's try #1942, if it won't work, we can reopen this one and merge it.

PR #1928, opened 2020-11-02
Fix #749 - Fix Walk if is hostile = false

### LINKED ISSUE #749: Bug "runonhealth" + "limit of run"
When the monsters starts to run, he get a few "steps" to move, and keep moving , like 1/2 second.
Also.. The "limit" of running is too close, it should be atleast 24~30 sqms from the player.

Is kinda simple to be explained, but i've produce a video to show it.

https://www.youtube.com/watch?v=inb61vaGANI

heberholanda:


debone (approved):
And with this one, isPassive idea dies :) Let's hope CipSoft doesn't release new monsters with such functionality and then we would need to have such

PR #1929, opened 2020-11-02
searchTarget update objects and switch cases

heberholanda:
- Standardizing objects
- Using the switch for default value
- Loop optimizations


src/monster.cpp
```diff
@@ -512,59 +512,56 @@ void Monster::onCreatureLeave(Creature* creature)
...
-				resultList.push_back(creature);
+	//lets just pick the first target in the list
+	for (Creature* target : targetList) {
+		if (followCreature != target && isTarget(target)) {
+			if (searchType == TARGETSEARCH_RANDOM || canUseAttack(myPos, target)) {
+				resultList.push_back(target);
 			}
 		}
+		if (followCreature != target && selectTarget(target)) {
+			return true;
```
debone, line 526: Can we have more insight why this return was added? If I read correctly, there's no way to the rest of the attack strategies to work.

src/monster.cpp
```diff
@@ -512,59 +512,56 @@ void Monster::onCreatureLeave(Creature* creature)
...
 				if (++it != resultList.end()) {
-					const Position& targetPosition = getTarget->getPosition();
-					int32_t minRange = std::max<int32_t>(Position::getDistanceX(myPos, targetPosition), Position::getDistanceY(myPos, targetPosition));
+					const Position& targetPosition = target->getPosition();
+					int32_t minRange = Position::getDistanceX(myPos, targetPosition) + Position::getDistanceY(myPos, targetPosition);
 					do {
 						const Position& pos = (*it)->getPosition();
 
-						int32_t distance = std::max<int32_t>(Position::getDistanceX(myPos, pos), Position::getDistanceY(myPos, pos));
+						int32_t distance = Position::getDistanceX(myPos, pos) + Position::getDistanceY(myPos, pos);
```
debone, line 562: Can you give more details on these changes from `std::max` to `+`?

src/monster.cpp
```diff
@@ -641,16 +638,15 @@ bool Monster::searchTarget(TargetSearchType_t searchType /*= TARGETSEARCH_DEFAUL
 				std::advance(it, uniform_random(0, resultList.size() - 1));
 				return selectTarget(*it);
 			}
+
+			if (searchType == TARGETSEARCH_ATTACKRANGE) {
```
debone, line 642: Hum, so if it is ranged attack it won't have a target?

PR #1800, opened 2020-10-01
Optimized Clean with optional Protection Zone Clean

heberholanda:
### Optimized Clean with optional Protection Zone Clean

**config.lua**
`cleanProtectionZones = false`

/clean command:
Screenshot_65

/clean command updated:
Screenshot_65OPT

**Code From TFS**
Link: https://github.com/otland/forgottenserver/pull/2647

PR #1930, opened 2020-11-03
Clean performance fix

heberholanda:
**Old PR:** 
Optimized Clean with optional Protection Zone Clean [#1800](https://github.com/opentibiabr/otservbr-global/pull/1800)

A bug was detected and the code was removed. I'm bringing a **_TFS_** fix made by **@TheSumm**

@yatozin Can test it ?

**@TheSumm**
**Thank you for that**


src/game.h
```diff
@@ -565,6 +565,19 @@ class Game
 		std::forward_list<Item*> toDecayItems;
 		std::forward_list<Item*> toImbuedItems;
 
+		std::unordered_set<Tile*> getTilesToClean() const {
```
SaiyansKing, line 568: It should return a reference " std::unordered_set<Tile*>& " because it'll keep making unnecessary copies every time getTilesToClean() gets called.

heberholanda: humm..

dex-89 (changes requested):
Check conflicts

dex-89:
I tested it. Works, but sometimes crash server `Memory protection violation`. I try fix it in other PR

PR #1931, opened 2020-11-03
remove query executing several times

heberholanda:
Thanks
@debone && @marcosvf132

### DIFF
--- a/src/spawn.cpp
+++ b/src/spawn.cpp
@@ -57,6 +57,10 @@ bool Spawns::loadFromXml(const std::string& fromFilename)
 	DBResult_ptr BoostedName = db.storeQuery(query.str());
 	std::string BoostedNameGet = BoostedName->getString("boostname");
 
+	if (!BoostedName) {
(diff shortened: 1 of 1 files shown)


src/spawn.cpp
```diff
@@ -93,11 +98,6 @@ bool Spawns::loadFromXml(const std::string& fromFilename)
 				
 				int32_t boostedrate;
 				
-				Database& db = Database::getInstance();
-				std::ostringstream query;
-				query << "SELECT `boostname` FROM `boosted_creature`";
-				DBResult_ptr BoostedName = db.storeQuery(query.str());
-				
 				if (!BoostedName) {
```
debone, line 101: The error is also part of the query :)

debone: And I'm not sure we want to fail monster load due lack of a boosted creature. Maybe just the warning and a sane default would be enough?

heberholanda: I remove return false, Thanks brow

debone (approved):

PR #1933, opened 2020-11-03
CorpseId fix

heberholanda:
- Dark Monk
- Monk
- Insect Swarm


dex-89 (approved):
Dead of insect swarm has duration ~3 seconds but great work :)

debone (approved):

PR #1934, opened 2020-11-03
Replace Hardcode

heberholanda:


src/player.cpp
```diff
@@ -3467,7 +3467,7 @@ void Player::internalAddThing(uint32_t index, Thing* thing)
 	}
 
 	//index == 0 means we should equip this item at the most appropiate slot (no action required here)
-	if (index > 0 && index < 12) {
+	if (index > CONST_SLOT_WHEREEVER && index <= CONST_SLOT_LAST) {
```
debone: To keep it consistent
```suggestion
	if (index >= CONST_SLOT_FIRST && index <= CONST_SLOT_LAST) {
```

debone (approved):
Soft reminder about the commit/PR messages :) "Replace Hardcode" can be more descriptive. You can also add some descriptions with "why" the changes

PR #1936, opened 2020-11-03
config.lua houseOwnedByAccount

heberholanda:
config: houseOwnedByAccount 

This enables or disables the possibility that characters from the same account automatically become "owners" and have full access to the house.

From TFS by @MillhioreBT


dex-89 (approved):
Works perfectly :) I tested as false and true in config.lua. Great work 👍

src/iologindata.cpp
```diff
@@ -81,6 +81,19 @@ uint32_t IOLoginData::gameworldAuthentication(const std::string& email, const st
   return accountId;
 }
 
+uint32_t IOLoginData::getAccountIdByPlayerName(const std::string& playerName)
+{
+	Database& db = Database::getInstance();
+
+	std::ostringstream query;
+	query << "SELECT `account_id` FROM `players` WHERE `name` = " << db.escapeString(playerName);
```
debone: Can't we look for player by their id? I'm scratching my head wondering if any crazy name wouldn't send this to space and it feels so much better to simply use the id?

debone (changes requested):

dex-89 (approved):
I have tested and the test is passed.

PR #1938, opened 2020-11-03
getDistanceBetween to Position:getDistanceBetween(position)

heberholanda:


data/lib/core/position.lua
```diff
@@ -110,6 +110,16 @@ function Position.getTile(self)
 	return Tile(self)
 end
 
+function Position:getDistanceBetween(position)
+	local xDif = math.abs(self.x - position.x)
+	local yDif = math.abs(self.y - position.y)
+	local posDif = math.max(xDif, yDif)
+	if self.z ~= position.z then
```
debone, line 117: I know that your PR is just moving the function around :P but there's a potential issue here where it assumes that 1 or 10 floors difference will be the same. And I'm not sure one floor difference amounts to 15 sqm :)

debone (approved):
I approved but consider my comment

PR #1944, opened 2020-11-07
Prompt wait for a key to close (if gets an error)

IgorCFG:
## Description (eng)
If the initialization has a bug the `system("pause")` will make the prompt wait for a key to close.
For me is useful, because I didn’t know what was causing the prompt to close as there wasn’t even time to see what was on it.

## Descrição (ptBr)
Se a inicialização do servidor tiver algum bug, o `system("pause")` irá fazer com que o prompt espere uma tecla para ser finalizado.
Pra mim é útil, porque eu não sabia o que estava causando o fechamento dele na inicialização, já que nem dava tempo de ler.

### DIFF
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -56,6 +56,7 @@ std::unique_lock<std::mutex> g_loaderUniqueLock(g_loaderLock);
 void startupErrorMessage(const std::string& errorStr) {
 	std::cout << "> ERROR: " << errorStr << std::endl;
 	g_loaderSignal.notify_all();
+  system("pause");
 }
 
 void mainLoader(int argc, char* argv[], ServiceManager* servicer);
--- a/src/otserv.cpp
+++ b/src/otserv.cpp
@@ -54,9 +54,11 @@ std::condition_variable g_loaderSignal;
 std::unique_lock<std::mutex> g_loaderUniqueLock(g_loaderLock);
 
 void startupErrorMessage(const std::string& errorStr) {
-	std::cout << "> ERROR: " << errorStr << std::endl;
-	g_loaderSignal.notify_all();
-  system("pause");
+  std::cout << "\033[1;31m>> " << errorStr << std::endl;
+  std::cout << ">> The program WILL CLOSE after receiving a key..." << "\033[0m" << std::endl;
+  g_loaderSignal.notify_all();
+  getchar();
+  exit(-1);
 }
 
 void mainLoader(int argc, char* argv[], ServiceManager* servicer);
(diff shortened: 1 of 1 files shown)


debone:
Thanks for your contribution @IgorCFG :) I can see how the idea of not exiting immediately after an error could benefit the users with less experience in debugging. It's common for these users to report issues and be helpless as "the window close too fast to know".

I didn't know about system calls and I did a short research on it. It seems that [system pause is Windows specific and not recommended practice](https://stackoverflow.com/questions/1107705/systempause-why-is-it-wrong). On the answers there are some suggestions to use the stdin to emulate such pause and wait for a enter press. I think it would be a nice way to introduce it instead of the system pause. 

Also, can you include a message before to let the user now that the server is going to terminate after enter press, due an unhandled error?

debone (changes requested):
(I have added a comment earlier)

IgorCFG:
> I didn't know about system calls and I did a short research on it.

Nice research! 🚀

I had done it with `getchar` but I think I ended up putting it in the wrong place (since it didn't work).
I read the comments and I found it interesting the part where you say that the `system("pause")` ends up being dependent on a file called "pause" and can also cause problems if for some reason you have a file with that name. 
Interesting, I'll make the adjustments.

Thanks!

IgorCFG:
@debone look the last commit there b8bb0f5

And also added color in the error message on `startupErrorMessage` method. 😎

## Demo
image

src/otserv.cpp
```diff
@@ -54,8 +54,11 @@ std::condition_variable g_loaderSignal;
 std::unique_lock<std::mutex> g_loaderUniqueLock(g_loaderLock);
 
 void startupErrorMessage(const std::string& errorStr) {
-	std::cout << "> ERROR: " << errorStr << std::endl;
-	g_loaderSignal.notify_all();
+  std::cout << "\033[1;31m>> " << errorStr << std::endl;
+  std::cout << ">> The program WILL CLOSE after receiving a key..." << "\033[0m" << std::endl;
```
debone: Maybe?
```suggestion
  std::cout << ">> Due an error, the program will close after pressing the enter key..." << "\033[0m" << std::endl;
```

IgorCFG: In some cases has more than one error message.
I don’t think it would take so many "there was an error", it ends up being redundant. 🤔

But the `receiveing a` for `pressing the enter` I liked it!
I thought it would close after any key's pressed 🤔

debone: ~~Will it repeat with the addition of~~ `exit(-1)` ~~?~~

Edit: Ach! I understood what you meant now. The error message might be extensive. All good :)

debone (approved):
Looks good now :) Keep in mind that your first commit looks like it was made with a work email. If you are ok with that, LGTM.

PR #1950, opened 2020-11-10
Cyclopedia -> Character

dex-89:
**I tested. It doesn't crash the client or the server. Works perfectly for me.**

**Before:**
2020-11-10_175247414_GOD_Hotkey

**After:**
2020-11-10_182556789_GOD_Hotkey

Compilation @ my Debian server:
screen

**Source code:** source from this [https://github.com/malucooo/otxserver-new]() project

PR #1954, opened 2020-11-11
Item inspection

dex-89:
**Tested and test passed**

function `getDescriptions` from item.cpp and item.h added too in #1950 

**Before:**
inspectionscreen
_not working..._

**After:**
2020-11-11_150959316_GOD_Hotkey

Compilation:
image

**No errors and crashed**


dex-89:
I must check it before

PR #1956, opened 2020-11-12
Refactoring - actions.cpp ♻

IgorCFG:
## Description
Looking to clean and improve the structure of the source, any comments to improve will be welcome.

## Adjustments
- Split `registerEvent` function into smaller functions (generic and specifics); ♻
- Created `enum ItemIdentifier_t` to identify the identifier being used (id, uid, aid); ♻
- Created `sendWarnLog` to send warning logs to prompt (this function will be changed by **spdlog** when implemented); ✨
- Created `@brief` for the new functions; ✨

## Note
I created on **DRAFT MODE** because it is not finished, but I would like to talk about these changes.

### Rebuild Log (updated on commit 27621c6)
```
1>------ Rebuild All started: Project: otservbr-global, Configuration: Release x64 ------
1>otpch.cpp
1>account.cpp
1>actions.cpp
1>ban.cpp
1>baseevents.cpp
1>bed.cpp
1>chat.cpp
1>combat.cpp
1>condition.cpp
1>configmanager.cpp
1>connection.cpp
1>container.cpp
1>creature.cpp
1>creatureevent.cpp
1>cylinder.cpp
1>database.cpp
1>databasemanager.cpp
1>databasetasks.cpp
1>depotchest.cpp
1>depotlocker.cpp
1>events.cpp
1>fileloader.cpp
1>game.cpp
1>gamestore.cpp
1>globalevent.cpp
1>groups.cpp
1>guild.cpp
1>house.cpp
1>housetile.cpp
1>imbuements.cpp
1>inbox.cpp
1>ioguild.cpp
1>iologindata.cpp
1>iomap.cpp
1>iomapserialize.cpp
1>iomarket.cpp
1>iostash.cpp
1>item.cpp
1>items.cpp
1>luascript.cpp
1>mailbox.cpp
1>map.cpp
1>modules.cpp
1>monster.cpp
1>monsters.cpp
1>mounts.cpp
1>movement.cpp
1>networkmessage.cpp
1>npc.cpp
1>otserv.cpp
1>outfit.cpp
1>outputmessage.cpp
1>party.cpp
1>player.cpp
1>position.cpp
1>protocol.cpp
1>protocolgame.cpp
1>protocollogin.cpp
1>protocolstatus.cpp
1>raids.cpp
1>reward.cpp
1>rewardchest.cpp
1>rsa.cpp
1>scheduler.cpp
1>script.cpp
1>scriptmanager.cpp
1>server.cpp
1>E:\@projects\tibia\otservbr-global\src\housetile.h(50,1): warning C4266: 'void Tile::internalAddThing(Thing *)': no override available for virtual member function from base 'Tile'; function is hidden (compiling source file ..\src\protocollogin.cpp)
1>E:\@projects\tibia\otservbr-global\src\tile.h(270): message : see declaration of 'Tile::internalAddThing' (compiling source file ..\src\protocollogin.cpp)
1>E:\@projects\tibia\otservbr-global\src\tile.h(153): message : see declaration of 'Tile' (compiling source file ..\src\protocollogin.cpp)
1>signals.cpp
1>spawn.cpp
1>spells.cpp
1>talkaction.cpp
1>tasks.cpp
1>teleport.cpp
1>thing.cpp
1>tile.cpp
1>tools.cpp
1>trashholder.cpp
1>vocation.cpp
1>waitlist.cpp
1>weapons.cpp
1>wildcardtree.cpp
1>xtea.cpp
1>Generating code
1>Finished generating code
1>otservbr-global.vcxproj -> E:\@projects\tibia\otservbr-global\vc14\x64\Release\otservbr-global-x64.exe
1>Done building project "otservbr-global.vcxproj".
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -177,7 +177,7 @@ bool Actions::registerChangeItemAction(Action* action, const pugi::xml_attribute
     sendWarnLog(log.c_str(), __FUNCTION__);
     return false;
   }
-  
+
   auto fromIdName = "from" + identifierName;
   auto fromItemId = pugi::cast<uint16_t>(attr.value());
   auto toItemId = pugi::cast<uint16_t>(toIdAttr.value());
@@ -206,8 +206,8 @@ bool Actions::registerChangeItemAction(Action* action, const pugi::xml_attribute
   return success;
 }
 
-void Actions::sendWarnLog(const char* log, const char* callMember) {
-  std::cout << "[Warning - " << callMember << "] " << log << std::endl;
+void Actions::sendWarnLog(const char* logMessage, const char* callMember) {
+  std::cout << "[Warning - " << callMember << "] " << logMessage << std::endl;
 }
 
 bool Actions::registerLuaEvent(Action* event)
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -119,8 +119,8 @@ bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node) {
 bool Actions::registerActionByItemId(Action* action, uint16_t itemId) {
   auto result = useItemMap.emplace(itemId, std::move(*action));
   if (!result.second) {
-    auto log = "Duplicate registered item with id : " + itemId;
-    sendWarnLog(log, __FUNCTION__);
+    auto log = "Duplicate registered item with id : " + std::to_string(itemId);
+    sendWarnLog(log.c_str(), __FUNCTION__);
   }
 
   return result.second;
@@ -129,8 +129,8 @@ bool Actions::registerActionByItemId(Action* action, uint16_t itemId) {
 bool Actions::registerActionByUniqueId(Action* action, uint16_t uniqueId) {
   auto result = uniqueItemMap.emplace(uniqueId, std::move(*action));
   if (!result.second) {
-    auto log = "Duplicate registered item with uniqueid : " + uniqueId;
-    sendWarnLog(log, __FUNCTION__);
+    auto log = "Duplicate registered item with uniqueid : " + std::to_string(uniqueId);
+    sendWarnLog(log.c_str(), __FUNCTION__);
   }
 
   return result.second;
@@ -139,8 +139,8 @@ bool Actions::registerActionByUniqueId(Action* action, uint16_t uniqueId) {
 bool Actions::registerActionByActionId(Action* action, uint16_t actionId) {
   auto result = actionItemMap.emplace(actionId, std::move(*action));
   if (!result.second) {
-    auto log = "Duplicate registered item with actionid : " + actionId;
-    sendWarnLog(log, __FUNCTION__);
+    auto log = "Duplicate registered item with actionid : " + std::to_string(actionId);
+    sendWarnLog(log.c_str(), __FUNCTION__);
   }
 
   return result.second;
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -153,7 +153,7 @@ bool Actions::registerActionByUniqueId(Action* action, uint16_t uniqueId)
 
 bool Actions::registerActionByActionId(Action* action, uint16_t actionId)
 {
-  auto result = uniqueItemMap.emplace(actionId, std::move(*action));
+  auto result = actionItemMap.emplace(actionId, std::move(*action));
   if (!result.second)
   {
     auto log = "Duplicate registered item with actionid : " + actionId;
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -83,44 +83,32 @@ Event_ptr Actions::getEvent(const std::string& nodeName)
 }
 
 
-bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node)
-{
-  //event is guaranteed to be an Action
+bool Actions::registerEvent(Event_ptr event, const pugi::xml_node& node) {
+  // event is guaranteed to be an Action
   Action_ptr actionPtr{ static_cast<Action*>(event.release()) };
 
   auto action = actionPtr.get();
 
   pugi::xml_attribute attr;
-  if ((attr = node.attribute("itemid")))
-  {
+  if ((attr = node.attribute("itemid"))) {
     auto itemId = pugi::cast<uint16_t>(attr.value());
     auto result = registerActionByItemId(action, itemId);
     return result;
-  }
-  else if ((attr = node.attribute("uniqueid")))
-  {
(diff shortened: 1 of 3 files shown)


src/actions.h
```diff
@@ -279,6 +279,59 @@ class Actions final : public BaseEvents
...
+    bool registerChangeItemAction(Action* action, const pugi::xml_attribute attr,
+                                  const pugi::xml_node& node, const ItemIdentifier_t& identifier);
+
+    /**
+     * @brief Send a warning log message to prompt
+     *
+     * @param logMessage message that will be sent
+     * @param callMember function that was at the time of the warning
+     */
+    void sendWarnLog(const char* logMessage, const char* callMember);
```
lgrossi, line 333: I personally deslike comments like that. They are extra lines that pollutes more than add value. 
Here the comment is only repeating things that are already in the code, nothing extra. 
Strongly typed languages such cpp are already clear enough if you praise for good function and variable names.

Don't get me wrong, comments are great, but i prefer to have them when you actually need: explain some logic that are not clear or some decisions that goes beyond the code. Otherwise i prefer it clean.

src/actions.cpp
```diff
@@ -82,120 +82,132 @@ Event_ptr Actions::getEvent(const std::string& nodeName)
...
+  // event is guaranteed to be an Action
+  Action_ptr actionPtr{ static_cast<Action*>(event.release()) };
+
+  auto action = actionPtr.get();
+
+  pugi::xml_attribute attr;
+  if ((attr = node.attribute("itemid"))) {
+    auto itemId = pugi::cast<uint16_t>(attr.value());
+    auto result = registerActionByItemId(action, itemId);
+    return result;
```
lgrossi, line 96: Why having result and not return the function itself?
return register...

src/actions.cpp
```diff
@@ -82,120 +82,132 @@ Event_ptr Actions::getEvent(const std::string& nodeName)
...
 
-		uint16_t fromAid = pugi::cast<uint16_t>(attr.value());
-		uint16_t iterAid = fromAid;
-		uint16_t toAid = pugi::cast<uint16_t>(toAidAttribute.value());
+  auto fromIdName = "from" + identifierName;
+  auto fromItemId = pugi::cast<uint16_t>(attr.value());
+  auto toItemId = pugi::cast<uint16_t>(toIdAttr.value());
+  auto iterId = fromItemId;
+  auto result = useMap.emplace(iterId, *action);
+  if (!result.second) {
```
lgrossi, line 186: Using auto everywhere is not a good practice. Strong typing is useful, clearer and makes reading easier.

src/actions.cpp
```diff
@@ -82,120 +82,132 @@ Event_ptr Actions::getEvent(const std::string& nodeName)
...
+    case ItemIdentifier_t::ID:
+      identifierName = "id";
+      useMap = useItemMap;
+      break;
+    case ItemIdentifier_t::UID:
+      identifierName = "uid";
+      useMap = uniqueItemMap;
+      break;
+    case ItemIdentifier_t::AID:
+      identifierName = "aid";
```
lgrossi, line 164: Instead of identifier name Id rather having a static function for your enum that gets the literal. Then you can use it more generically and centralised, without being a hardcoded conditional variable. Or maybe use const strings instead of enum directly. const string UID = "uid";

src/actions.h
IgorCFG: I don't like comments in general.
But in this case, I put it here for a possible generation of documentation in the future.
I'm prefer always a function name or header who's explains itself.

src/actions.cpp
IgorCFG: It's easier to debug.

lgrossi: You can just use debug tools to inspect values and variables, or either debug the result of this function, this is extra lines that you dont need imho.

src/actions.h
lgrossi: Well the function name should be clear enough about what does it do imho

src/actions.cpp
IgorCFG: Nice tip.

In C# I usually use `var`. 
You can read the type if u put the mouse pointer over, but It gets faster to know the type of the object if the type is on the side really, I will change it later.

src/actions.cpp
IgorCFG: I don't like static objects, they are left in memory unnecessarily, local objects are weaker and disposable, leaving the memory cleaner.

src/actions.cpp
IgorCFG: In that case, u need to go to next function to get the value, it's much more easier put a breakpoint on `return result` and get the `result` value.

IgorCFG: I prefer not to do `nested functions`, it's more readable if you separate the calls.

Maybe we can change the `result` name to some more specific.

src/actions.h
IgorCFG: How I said I wrote this to make the documentation with generator later.
Has some others @brief in that class.

I talked with @MajestyOTBR about that before,
If it were for me I would remove all this briefs and improve the function names (if it's needed) hehe

src/actions.cpp
lgrossi: This is not nested function lol

src/actions.cpp
```diff
@@ -82,120 +82,132 @@ Event_ptr Actions::getEvent(const std::string& nodeName)
...
+    auto actionId = pugi::cast<uint16_t>(attr.value());
+    auto result = registerActionByActionId(action, actionId);
+    return result;
+  } else if ((attr = node.attribute("fromid"))) {
+    auto result = registerChangeItemAction(action, attr, node, ItemIdentifier_t::ID);
+    return result;
+  } else if ((attr = node.attribute("fromuid"))) {
+    auto result = registerChangeItemAction(action, attr, node, ItemIdentifier_t::UID);
+    return result;
+  } else if ((attr = node.attribute("fromaid"))) {
```
lgrossi, line 111: Another thing, you're always returning you dont need those nested else ifs, you can split it in individual ifs

src/actions.cpp
IgorCFG: Ops, so ignore it, I'm lousy in theoretical terms. 😅
But my "point" continues, I think it's more readable to make a variable receive the function and not put a function inside the function parameter, in that case I'll change the name `result` to a better name, should improve.

src/actions.cpp
IgorCFG: Nice! 
Sounds better to read.

PR #1962, opened 2020-11-14
Protocol 12.52

dex-89:
**I am not sure if the line in `protocolgame.cpp` is in the right place! (add comment 12.52)**

**I think this is a change for new Linux users only.**

**Source:** [https://github.com/SaiyansKing/optimized_forgottenserver/commit/0e87c22e0af7b225bf46916ab3f4c0859a01dc8a](url)

**Windows client download:** [https://drive.google.com/file/d/1eCN_8P0z2OSjmJfpMQZXhWnfQh9xXKVS/view?usp=sharing](url)

**We need check and tests**

image


dex-89:
Goto 12.60

PR #1967, opened 2020-11-15
Many words in talkactions

heberholanda:
Support for multiple word triggers to a talkaction. 

**Revscriptsys:**
`local talk = TalkAction("/pos", "!pos")`

From TFS Credits **@rookgaard**


dex-89 (approved):
After change source command !pos and /pos works correctly and returns my position. Great work.

PR #1976, opened 2020-11-16
Yana cap check with amount

heberholanda:


dex-89 (approved):
I checked this. Very nice job! Works correct

PR #1982, opened 2020-11-16
Fix Houses/furniture tab on gamestore

cys9211:
parent = "House",   - We do not have such a tab in the store
We need to change to "Houses"


heberholanda (approved):
ok.

PR #2004, opened 2020-11-22
Remove misspell check

Costallat:
Removes misspell check as it is not helping us too much, there are some words that it complains that are false positive(correct words) and others wrong that are false negative(wrong words).

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -26,13 +26,6 @@ jobs:
       run: |
         mkdir -p $HOME/bin && curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh| sh -s -- -b $HOME/bin
 
-    - name: misspell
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewdog.yml
```diff
@@ -26,13 +26,6 @@ jobs:
       run: |
         mkdir -p $HOME/bin && curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh| sh -s -- -b $HOME/bin
 
-    - name: misspell
-      uses: reviewdog/action-misspell@master
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
-        locale: "US"
```
debone, line 33: I saw those errors and I thought they were because of UK vs US english. Maybe we could experiment with removing locale before removing the whole thing?

dudantas (approved):

PR #2011, opened 2020-11-25
Fixed NPC Spawn

dex-89:
Fixed NPC spawn after period change. For example: period change was xx:15 and server started xx:17

### DIFF
--- a/data/startup/others/functions.lua
+++ b/data/startup/others/functions.lua
@@ -110,9 +110,25 @@ function loadLuaMapBook(tablename)
 end
 
 function loadLuaNpcs(tablename)
+if (tonumber(os.date("%M", os.time())) < 15 or tonumber(os.date("%M", os.time())) >= 45) then
+	dayornight="night"
+elseif (tonumber(os.date("%M", os.time())) >= 15 or tonumber(os.date("%M", os.time())) < 45) then
+	dayornight="day"
+end
 	for index, value in pairs(tablename) do
 		if value.name and value.position then
-			local spawn = Game.createNpc(value.name, value.position)
+			local spawn
+			if value.spawn~=nil then
+				if dayornight=="day" and value.spawn=="day" then
+					spawn = Game.createNpc(value.name, value.position)
+					print("> NPC " .. value.name .. " added!")
+				elseif dayornight=="night" and value.spawn=="night" then
+					spawn = Game.createNpc(value.name, value.position)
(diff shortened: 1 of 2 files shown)


debone:
I think you need to reconsider the approach to this one. The periodChange is not hard-coded to xx:15 or xx:45, it's based on the time you started the server. If you start at xx:43, the first night will start at xx:58

data/startup/tables/npc_spawn.lua
```diff
@@ -493,7 +493,7 @@ NpcTable = {
 	[492] = {name = "willem", position = {x = 32063, y = 32020, z = 7}},
 	[493] = {name = "neill", position = {x = 32076, y = 32027, z = 7}},
 	[494] = {name = "lesser_messenger_of_heaven", position = {x = 32536, y = 32026, z = 7}},
-	[495] = {name = "ghostly_wolf", position = {x = 33332, y = 32051, z = 7}},
+	[495] = {name = "ghostly_wolf", position = {x = 33332, y = 32051, z = 7}, spawn="night"},
```
debone, line 496: Maybe spawnPeriod?

data/startup/others/functions.lua
```diff
@@ -110,9 +110,25 @@ function loadLuaMapBook(tablename)
...
+				if dayornight=="day" and value.spawn=="day" then
+					spawn = Game.createNpc(value.name, value.position)
+					print("> NPC " .. value.name .. " added!")
+				elseif dayornight=="night" and value.spawn=="night" then
+					spawn = Game.createNpc(value.name, value.position)
+					print("> NPC " .. value.name .. " added!")
+				end
+			else
+				spawn = Game.createNpc(value.name, value.position)
+			end
```
debone, line 131: I think you can simply check the server period when spawning, and if it matches the spawnPeriod of the NPC. You don't need all this logic nor OS time.

dex-89: @debone I am not good at lua. For me, the simplest solutions are the best. As for the period, I tested it all evening yesterday. 1 dropped out a minute earlier, the second dropped out 5 minutes later and 3 was gone Despite the time of day change (the globe in the Tibia client top right), there were still night NPCs.
If you have the will and knowledge to do it differently, I am open to suggestions. I am often not only here but also on 'discord'. We can always arrange an interview :)

majestyotbr:
@dudantas PR waiting for your review.

PR #2013, opened 2020-11-26
Fix: Monsters custom spells

marcosvf132:
Custom spells on data/spells that didn't have default damage wasn't dealing any damage at all.
Example: Aftershock (aftershock wave)

Before this PR:
before

With this PR:
after


dudantas (approved):
Thanks for your contribution

PR #2020, opened 2020-11-27
Revert "Adding things (map and otb) for compatibility with version 12.40/12.55 and fix bugs"

dex-89:
Reverts opentibiabr/otservbr-global#2012 @dudantas 

image

**Errors:**
`Duplicate unique id: 64142`
`Memory protection violation`


majestyotbr:
I don't approve this revert, it's better to fix the problem.

PR #2024, opened 2020-11-28
quickloot_fallback is added in migration file

dex-89:
In file `data/migrations/5.lua` is `db.query("ALTER TABLE `players` ADD `quickloot_fallback` TINYINT DEFAULT 0")`. That line has error:
```
> Updating database to version 6 (quickloot)
[Error - mysql_real_query] Query: ALTER TABLE `players` ADD `quickloot_fallback` TINYINT DEFAULT 0
Message: Duplicate column name 'quickloot_fallback'
```

### DIFF
--- a/schema.sql
+++ b/schema.sql
@@ -152,7 +152,6 @@ CREATE TABLE IF NOT EXISTS `players` (
   `marriage_status` bigint(20) UNSIGNED NOT NULL DEFAULT '0',
   `marriage_spouse` int(11) NOT NULL DEFAULT '-1',
   `bonus_rerolls` bigint(21) NOT NULL DEFAULT '0',
-  `quickloot_fallback` tinyint(1) DEFAULT '0',
   INDEX `account_id` (`account_id`),
   INDEX `vocation` (`vocation`),
   CONSTRAINT `players_pk` PRIMARY KEY (`id`),


schema.sql
```diff
@@ -152,7 +152,6 @@ CREATE TABLE IF NOT EXISTS `players` (
   `marriage_status` bigint(20) UNSIGNED NOT NULL DEFAULT '0',
   `marriage_spouse` int(11) NOT NULL DEFAULT '-1',
   `bonus_rerolls` bigint(21) NOT NULL DEFAULT '0',
-  `quickloot_fallback` tinyint(1) DEFAULT '0',
```
dudantas, line 155: I believe that the correct thing is to check if the table already exists to be able to perform the migration. Because it is ideal to keep the schema always up to date, the function of migrations is to update the database of those who already have an online server, the schema is used for new installations.

dex-89: But i used `schema.sql` with current repository to this tests

dudantas: Yes. So I think that the correct thing is for migrations to check if the table already exists, and if there is ignore.

dex-89: Is ignore but with error in console :) But ok 👍

dudantas: You did not understand. I mean we have to add a function that checks if the table already exists in the database, and if it already exists, it will ignore migrations (thus avoiding the error on the console)

PR #2026, opened 2020-11-28
Tibia raining with thunder

dex-89:
Have you seen the rain in your Tibia? Look this :)


aurelion5670:
Why this should be added if that isn't a feature of lastest tibia?

xThauma:
> Why this should be added if that isn't a feature of lastest tibia?

It will probably be added in the next summer update anyway.

dex-89:
> > Why this should be added if that isn't a feature of lastest tibia?
> 
> It will probably be added in the next summer update anyway.

I doubt it. This is for diversifying the Tibia content

debone:
I think while this is great, it has no space in here. Yes, there’s a lot of room to improvements and diversification for Tibia, but then, where should we stop? You can and should definitely make this content available in other places for other otserv to use it, but I don’t think merging here is the way.

PR #2042, opened 2020-12-03
Fix a Familiars summons

dex-89:
```
Characteristics
Their speed is the same as their master's.
If they get too far behind from their master, they teleport themselves to him/her.
They automatically go up and down ramps, stairs, etc., as well as travel together when their master uses any travel
system or teleport.
When they are not in battle, their health regenerates very quickly.
They only last for 15 minutes after being summoned. This timer does not count when the character is offline.
They are not allowed in fights against Lever Bosses.
```

https://tibia.fandom.com/wiki/Familiars

### DIFF
--- a/data/lib/core/storages.lua
+++ b/data/lib/core/storages.lua
@@ -134,6 +134,7 @@ Storage = {
 	StoreExaust = 30051,
 	LemonCupcake = 30052,
 	BlueberryCupcake = 30053,
+	PetSummonTime = 30054,
 	PremiumAccount = 998899,
 	
 	--[[
--- a/data/scripts/creaturescripts/others/summon.lua
+++ b/data/scripts/creaturescripts/others/summon.lua
@@ -1,62 +1,88 @@
-local summon = {
-	[VOCATION.ID.MASTER_SORCERER] = {name = "thundergiant"},
-	[VOCATION.ID.ELDER_DRUID] = {name = "grovebeast"},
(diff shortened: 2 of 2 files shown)


brunomaidana97:
if your summon dies, just relog him to come back alive (still counting the time he had when he died)

19:08 You see a skullfrost. (Master: Knight Sample). It will disappear in 4 minutes and 30 seconds. < this is after i relogged and he appeared

after 15 min he doesn't disappear.

19:17 You see a skullfrost. (Master: Knight Sample). It will disappear in -5 minutes and 23 seconds.

another thing I would like to complement, sometimes the summon disappears during the hunt when there are many creatures, apparently the creatures push him, or he tries to go over, I could not identify. I did some tests and when changing this line pushable = true to false (it's inside the summon file) it didn't happen anymore. some testing is needed.

dex-89:
> if your summon dies, just relog him to come back alive (still counting the time he had when he died)
> 
> 19:08 You see a skullfrost. (Master: Knight Sample). It will disappear in 4 minutes and 30 seconds. < this is after i relogged and he appeared
> 
> after 15 min he doesn't disappear.
> 
> 19:17 You see a skullfrost. (Master: Knight Sample). It will disappear in -5 minutes and 23 seconds.

Thanks for information. Tomorrow i will be make rework and try fix problem

dex-89:
thanks @debone for paying attention

PR #2053, opened 2020-12-07
Fixed: Monsters stepdancing and place creature in field 

dex-89:
- **Monsters can now stepdancing in fields when attacking a foe.**
- **Monsters are now ignoring fields in the ground when they are created.**

**Author:** @FakeShinoda

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -231,6 +231,7 @@ bool Map::placeCreature(const Position& centerPos, Creature* creature, bool exte
 			if (monster) {
 				monster->ignoreFieldDamage = false;
 			}
+		}
 	}
 
 	int32_t index = 0;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -213,10 +213,10 @@ bool Map::placeCreature(const Position& centerPos, Creature* creature, bool exte
 				continue;
 			}
 
-      if (monster) {
+			if (monster) {
 				monster->ignoreFieldDamage = true;
 			}
-      
+
 			if (tile->queryAdd(0, *creature, 1, FLAG_IGNOREBLOCKITEM | FLAG_IGNOREFIELDDAMAGE) == RETURNVALUE_NOERROR) {
 				if (!extendedPos || isSightClear(centerPos, tryPos, false)) {
 					foundTile = true;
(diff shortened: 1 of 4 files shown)


src/monster.cpp
```diff
@@ -1886,7 +1886,7 @@ bool Monster::canWalkTo(Position pos, Direction moveDirection) const
 		}
 
 		Tile* walkTile = g_game.map.getTile(pos);
-		if (walkTile && walkTile->getTopVisibleCreature(this) == nullptr && walkTile->queryAdd(0, *this, 1, FLAG_PATHFINDING) == RETURNVALUE_NOERROR) {
+		if (tile && tile->getTopVisibleCreature(this) == nullptr && tile->queryAdd(0, *this, 1, FLAG_PATHFINDING | FLAG_IGNOREFIELDDAMAGE) == RETURNVALUE_NOERROR) {
```
gerotib, line 1889:
```suggestion
		if (walkTile && walkTile->getTopVisibleCreature(this) == nullptr && walkTile->queryAdd(0, *this, 1, FLAG_PATHFINDING | FLAG_IGNOREFIELDDAMAGE) == RETURNVALUE_NOERROR) {
```
should not be changed

FakeShinoda: replace walkTile to tile

src/monster.cpp
```diff
@@ -1886,7 +1886,7 @@ bool Monster::canWalkTo(Position pos, Direction moveDirection) const
 		}
 
 		Tile* walkTile = g_game.map.getTile(pos);
```
gerotib:
```suggestion
		Tile* tile = g_game.map.getTile(pos);
```

gerotib:
Monsters are walking on fields all the time with this change, not only while dancing.

dex-89:
Maybe this: https://otland.net/threads/monster-walk-field.246167/ will help something?

FakeShinoda:
> 
> 
> Monsters are walking on fields all the time with this change, not only while dancing.

Right, thanks for the report. <3

src/map.cpp
```diff
@@ -215,7 +227,10 @@ bool Map::placeCreature(const Position& centerPos, Creature* creature, bool exte
 
 		if (!foundTile) {
 			return false;
-		}
+		} else {
+			if (monster) {
+				monster->ignoreFieldDamage = false;
+			}
```
gerotib, line 233:
```suggestion
			}
		}
```
Missing here.

gerotib (commented):
With this small fix this pr seems to work fine.

gerotib (approved):

FakeShinoda (approved):

PR #2065, opened 2020-12-11
Fixed market in event room

dex-89:
**Maked new city 'Event room' and set DepotID as it**


dudantas (approved):
Thanks for your contribution

PR #2068, opened 2020-12-12
Cyclopedia -> Houses

dex-89:
**Need reviews, checks, tests**

**Who can help me?**

**Author:** @SaiyansKing

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -1256,7 +1256,7 @@ void ProtocolGame::parseCyclopediaHouseAction(NetworkMessage& msg)
 	//(void)msg;
 	uint8_t houseActionType = msg.getByte();
 	switch (houseActionType) {
-		case 0: { //wybor domku?
+		case 0: { // select city?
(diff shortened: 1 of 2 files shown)


aurelion5670:
This is the base code for houses of saiayans, it's pretty incomplete and i wouldn't recommend it to merge until it's done (As isn't done at all)

dex-89:
> This is the base code for houses of saiayans, it's pretty incomplete and i wouldn't recommend it to merge until it's done (As isn't done at all)

I know that, I'm trying to finish it.

SaiyansKing:
It isn't even code for houses, I only tested the cyclopedia houses structure. The reason because I didn't even touched it after so much time is because I don't create another real map server so I need to code program(add it to rme?) that will create client map files with house details.

FakeShinoda:
I started this system with a modification in RME to include the house details and parse to server.
If you can wait a little bit i can help you to finished that... But like they say its just the houses structure.

Issue #2029, opened 2020-11-29
BUG: Necromancer/Minotaur task count dont appears in quest log


brunomaidana97:
#### Describe the bug
> when taking the minotaur task the counter does not appear in the questlog

> Give us all details to reproduce the behavior:
> 1. Go to npc Budrik (task minotaurs)
> 2. say hi > mission > yes > task > yes
> 3. Your questlog has been updated.
> 4. Open your questlog, enter in Killing in the name of...
> 5. there the minotaurs task should appear.

#### Expected behavior
> like this (example)
![c0apq1h](https://user-images.githubusercontent.com/64701283/100529000-939e3400-31c1-11eb-8621-fedf9db43bd1.png)

#### Additional context
> you also need to check the task with npc Lugri (task necromancer), apparently the same bug is happening.

PR #2106, opened 2020-12-22
Fix Necro/Mino task quest log

hyresu:
Fix for #2029


dudantas (approved):
Thanks for your contribution!

PR #2109, opened 2020-12-22
Fix summons bug

dudantas:
Fixed function that, if the player was premium, returned false (missing "not")
Identified a few lines
Modified from checking the vocation of ID for CLIENT_ID
Thanks @leotk4


LeoTKBR (changes requested):
```
    if not player or not isPremium(player) then
		player:getPosition():sendMagicEffect(CONST_ME_POFF)
		player:sendCancelMessage("You need a premium account.")
        return false
    end
```

dudantas:
> ```
>     if not player or not isPremium(player) then
> 		player:getPosition():sendMagicEffect(CONST_ME_POFF)
> 		player:sendCancelMessage("You need a premium account.")
>         return false
>     end
> ```

Thanks

Issue #2048, opened 2020-12-06
BUG: Ferumbras' Ascendant Quest Access


cys9211:
#### Describe the bug
> After talking with npc Mazarius, we take missions and give back 30 demonic essences, after which the teleport does not take us to a location

> Give us all details to reproduce the behavior:
> 1. Talk Mazarius
> 2. Give 30 demonic essence 
> 3. Cant go to teleport

#### Environment
- OS (e.g. Windows): Windows 10
- OTServ Version (e.g. commit): 1.0 
- Tibia Version (e.g. 12.52): 12.52

PR #2113, opened 2020-12-22
Fix for Ferumbras Ascendant Entrance

hyresu:
Fix for #2048 

Also @dudantas I think you are making the otbm pulls, right? I think the teleport at Pos(33317, 32315, 13) should only have the action id, as you can see it teleports you as well activating the move event, thx in advance
image


dudantas (approved):

dudantas:
> Fix for #2048
> 
> Also @dudantas I think you are making the otbm pulls, right? I think the teleport at Pos(33317, 32315, 13) should only have the action id, as you can see it teleports you as well activating the move event, thx in advance
> image

So have to remove the destination from the teleport, correct?

majestyotbr (approved):

PR #2108, opened 2020-12-22
Feat: Map corrections

dudantas:
Minor corrections in respawns and corrections in the general map, added items where none existed, corrected some houses and removed items that were not needed
Fixed error in loadLuaMapAction (variable with wrong name)

Registered new doors, ids: 
38467
38468
38469
38470
38471
38472
39056
30057
39058
39059

Fixed wooden ladder from id 35964 and 35965

Thanks for @dex-89

PR #2133, opened 2020-12-23
Fixed house prices (missing in #2108)

dex-89:


dudantas:
You need to change the houses directly in the otbm file, only XML is useless (it only serves to store information.)

dex-89:
@dudantas you finished edit map?

dudantas:
> @dudantas you finished edit map?

Yes Sir

dudantas:
Added to https://github.com/opentibiabr/otservbr-global/pull/2261

PR #2146, opened 2020-12-24
Fix sale of golden outfit through the store

LeoTKBR:
Now it will be possible to win the decoration of the house and the storage of the golden outfit thus avoiding the duplicate purchase of the outfit and the fact of not winning the decoration of the house if you buy the outfit through the store


dudantas:
Thank you very much for your contribution.
I tested it and had a problem, it crashed the client.
However, anyway onLogin is not the best way (for having to relog to be able to work)
I'll put it to run inside the store, when buying the item. However, I am very grateful because it was thanks to your PR that I had this idea.

PR #2190, opened 2020-12-29
Fix Rashid's daily travel routine

fernandomatos:
## Case

Recently I realized that Rashid was being too lazy on my server, I never found him in other places, besides where I saw him for the first time.

Well, at least in my case, he never travels. I tried to figure out if it had something coded in the *save server files* handling this job, but I really haven't found anything in there related or in any other place. 

## Fix

To solve this, I modified `scripts/globalevents/spawn/rashid.lua`  inserting a `GlobalEvent` named "*rashid spawn*", and based on the other project files, I call two methods:

### onStartup()
First, I'm using **onStartup()** to create and position Rashid at the correct position using the array to get the position for the current day of the week. 

### onTime()
After that, I'm using **onTime()** to finally stop Rashid's laziness and put him to travel at 07:00 o'clock in the morning, every day.

Now he travels with the server running without any kind of restarts/reloads to all the cities once a day.

### Modification in additional_map.lua
I noticed that there is a custom map enabled for additional loading, with all important NPCs, probably to some kind of test. To avoid having a second Rashid in the map, I set it to false to disable additional map loading with all these extra NPCs.

Btw, I'm studying LUA and getting to know the organization of your project. Forgive me for some of my mistakes or if my approach is not good, I have other tech backgrounds and through them, I'm trying to learn the type of code you are using here, to contribute with the project in more ways.

If you could point out eventually some mistake that I made or a better approach to do this routine, I will be glad to learn with. Thank you!

### DIFF
--- a/data/startup/tables/additional_map.lua
+++ b/data/startup/tables/additional_map.lua
@@ -4,6 +4,6 @@ CustomMapTable = {
 		mapName = "otservbr-custom",
 		mapFile = "data/world/custom/otservbr-custom.otbm",
 		spawnFile = "data/world/custom/otservbr-custom-spawn.xml",
-		enabled = true
+		enabled = false
 	}
 }
--- a/data/startup/tables/additional_map.lua
+++ b/data/startup/tables/additional_map.lua
@@ -4,6 +4,6 @@ CustomMapTable = {
 		mapName = "otservbr-custom",
 		mapFile = "data/world/custom/otservbr-custom.otbm",
 		spawnFile = "data/world/custom/otservbr-custom-spawn.xml",
-		enabled = false
+		enabled = true
 	}
 }
--- a/data/scripts/globalevents/spawn/rashid.lua
+++ b/data/scripts/globalevents/spawn/rashid.lua
@@ -1,6 +1,3 @@
-local today = os.date("*t").wday
-local todayLabel = os.date("%A")
-
 local positionByDay = {
 	[1] = Position(32328, 31782, 6), -- Sunday
 	[2] = Position(32207, 31155, 7), -- Monday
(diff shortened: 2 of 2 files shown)


dex-89 (approved):
Avoid duplicating the `rashidSpawn` variable names. It looks perfect after fix

dex-89 (commented):
`Recently I realized that Rashid was being too lazy on my server, I never found him in other places, besides where I saw him for the first time.`

`Well, at least in my case, he never travels. I tried to figure out if had something coded in the save server files for handling this job, but I really haven't found anything in there related or in any other place else.`

Rashid travels after saving and rebooting the server. If you don't restart the server your fix is very good

fernandomatos:
Thank you @dex-89 for reviewing and teaching me more about this scenario ! It's excellent to know that I can use camel case naming convention for global event names and also that it is a good practise to separate global events, in single responsibility variables.

I will do this way on the next ones. I appreciate your review and the time you dedicated replying to me.

dex-89:
Join to us at Discord: https://discord.com/invite/3NxYnyV

data/startup/tables/additional_map.lua
```diff
@@ -4,6 +4,6 @@ CustomMapTable = {
 		mapName = "otservbr-custom",
 		mapFile = "data/world/custom/otservbr-custom.otbm",
 		spawnFile = "data/world/custom/otservbr-custom-spawn.xml",
-		enabled = true
+		enabled = false
```
dudantas: You just need to reactivate the loading of the custom map.

dex-89: Yes, but if Rashid is on a custom map, it won't be deleted and moved to another location by script?

fernandomatos: Thanks @dudantas for pointing this out. I think @dex-89 is right, because when the script searches and teleports Rashid by the name, it will teleport the first one found, causing a side effect of two Rashids on the main map. 

Is it necessary we have a Rashid spawn on the custom map? If it is, I can think about another approach considering that could be more Rashids on the map.

dudantas: It is not necessary. The Rashid from custom map could have another name, to prevent this problem from happening.

I'm going to pull a pull up fixing this custom Rashid issue, okay? There warning to be tested and we can merge this.

dudantas: Test in: https://github.com/opentibiabr/otservbr-global/pull/2279
If it's ok, let me know that I merge ...

dex-89: Everything will work fine with the `custom Rashid`. The script will find the `correct Rashid` and only move him to a new location.

dudantas: @fernandomatos you can set enable for true? Then we can merge.

dex-89 (approved):

data/startup/tables/additional_map.lua
fernandomatos: @dudantas Yes, for sure, thank you for making a custom Rashid to help this implementation to work properly.

dudantas (approved):

majestyotbr (approved):

PR #2186, opened 2020-12-28
round down required mana

gpedro:
cherry-pick from:
https://github.com/otland/forgottenserver/pull/2837

PR #2209, opened 2020-12-31
fix: vocations mana/skills

gpedro:
suggested by: DkO

### DIFF
--- a/data/XML/vocations.xml
+++ b/data/XML/vocations.xml
@@ -10,7 +10,7 @@
 		<skill id="5" multiplier="1.5" />
 		<skill id="6" multiplier="1.1" />
 	</vocation>
-	<vocation id="1" clientid="3" name="Sorcerer" description="a sorcerer" gaincap="10" gainhp="5" gainmana="30" gainhpticks="12" gainhpamount="1" gainmanaticks="3" gainmanaamount="2" manamultiplier="1.1" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="1">
+	<vocation id="1" clientid="3" name="Sorcerer" description="a sorcerer" gaincap="10" gainhp="5" gainmana="30" gainhpticks="12" gainhpamount="1" gainmanaticks="3" gainmanaamount="2" manamultiplier="1.2" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="1">
 		<formula meleeDamage="1.0" distDamage="1.0" defense="1.0" armor="1.0" />
 		<skill id="0" multiplier="1.5" />
 		<skill id="1" multiplier="2.0" />
@@ -20,7 +20,7 @@
 		<skill id="5" multiplier="1.5" />
 		<skill id="6" multiplier="1.1" />
 	</vocation>
-	<vocation id="2" clientid="4" name="Druid" description="a druid" gaincap="10" gainhp="5" gainmana="30" gainhpticks="12" gainhpamount="1" gainmanaticks="3" gainmanaamount="2" manamultiplier="1.1" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="2">
+	<vocation id="2" clientid="4" name="Druid" description="a druid" gaincap="10" gainhp="5" gainmana="30" gainhpticks="12" gainhpamount="1" gainmanaticks="3" gainmanaamount="2" manamultiplier="1.2" attackspeed="2000" basespeed="220" soulmax="100" gainsoulticks="120" fromvoc="2">
 		<formula meleeDamage="1.0" distDamage="1.0" defense="1.0" armor="1.0" />
 		<skill id="0" multiplier="1.5" />
 		<skill id="1" multiplier="1.8" />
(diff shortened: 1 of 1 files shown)


ElTeixe:
From what I been checking last days, playing/testing on Official and fixing dawnport stuffs and implementing magic level and skills conversion for vocations, some of that changes are wrong.

- The change of match dawnport vocations with the main vocations is correct, but they should keep the same hp/mana/capacity gain than none vocation 5/5/10.

- The change on the melee skills (0, 1, 2, 3) multipliers on paladin, royal paladin, dawnport paladin, knight, elite knight and dawnport knight is wrong. They must remain as it they before.

- The multipliers that really changed was the manamultipliers. I still working on figure out the new ones, but at least for sure knight vocations should be 4.0 instead of 3.0.

I still working to verify them all, but may take a while. If you want fix it here good, anyway I'll continue to work on my fix.

gpedro:
just in case, maybe we need to test this fix with PR #2186.

DeCarvalhoJr:
> From what I been checking last days, playing/testing on Official and fixing dawnport stuffs and implementing magic level and skills conversion for vocations, some of that changes are wrong.
> 
> * The change of match dawnport vocations with the main vocations is correct, but they should keep the same hp/mana/capacity gain than none vocation 5/5/10.
> * The change on the melee skills (0, 1, 2, 3) multipliers on paladin, royal paladin, dawnport paladin, knight, elite knight and dawnport knight is wrong. They must remain as it they before.
> * The multipliers that really changed was the manamultipliers. I still working on figure out the new ones, but at least for sure knight vocations should be 4.0 instead of 3.0.
> 
> I still working to verify them all, but may take a while. If you want fix it here good, anyway I'll continue to work on my fix.

Based on Wiki and by going to official tibia you'll see Paladin skill distance faster than Knight level up melee, so making both equal is incorrect..
Ofc it needs more testing but assuming they are equal is wrong.. 
paladin distance 1.1 
knight melee 1.2
isnt the correct due test but it fits to what official test results delivered and even what wiki/fandom says
Paladin level up distance faster than knight level up melee

`Avança sua habilidade de combate à distância mais rapidamente do que qualquer outra vocação. (Mais rápido que um Knight avança em combate corpo-a-corpo);`
[Tibiakiwi about paladin](https://www.tibiawiki.com.br/wiki/Paladin)

And since paladin skill melee slower than knight it has 1.3 for melee skills.. as said, wasnt made any test on it to have the correct values yet they fit with official 

```
None
ML 2 - 76.81%
Melee/Distance 12 - 22.50%```
```
Druid
ML 4 - 1.20%
Melee 13 - 98.97%
Distance 13 - 98.86%```
```
Sorcerer
ML 4 - 1.20
Melee/Distance 12 - 22.50% (same as none)```
```
Knight
ML 2 - 76.81% (same as none)
Melee 15 - 100%
Distance 13 - 36.15```
```
Paladin
ML 3 - 37.04%
Melee 14 - 64.43%
Distance 15 - 100%```

This result was made w/o any leveling skill just base one, as said reading wiki/fandom talls one thing, tests need to be done.. skill changes were made based on what was gotten from what was supposed to be

ElTeixe:
> just in case, maybe we need to test this fix with PR #2186.

Thats what Im doing, due that I suggested that change.

ElTeixe:
@TheAncientDragon 

You are confusing the stuffs...

On official results:
None vocation club, axe, sword 12 - 77.50% -> knight 15 00%
None vocation distance 12 - 77.50% -> paladin 15 00%

So the skill multipliers for club, axe, sword on knights and distance on paladins must be the same: 1.1

Paladin:
`
<skill id="4" multiplier="1.1" />
`

Knight:
`
<skill id="1" multiplier="1.1" />
<skill id="2" multiplier="1.1" />
<skill id="3" multiplier="1.1" />
`

And it was right before.


The only true fact about, is that paladins skill faster club, axe, sword than knights skill distance. 

On Official results:
None vocation distance 12 - 77.50% -> knight 13 - 63.85%
None vocation club, axe, sword 12 - 77.50% -> paladin 14 - 35.57%

Paladin:
`
<skill id="1" multiplier="1.2" />
<skill id="2" multiplier="1.2" />
<skill id="3" multiplier="1.2" />
`

Knight:
`
<skill id="4" multiplier="1.4" />
`

And it was right before too.

mrhultin:
> Based on Wiki and by going to official tibia you'll see Paladin skill distance faster than Knight level up melee, so making both equal is incorrect..

The reason for this, is that Blood hits/sparks (attacks where damage is fully absorbed by the monsters armor) are counted twice, not because their base multiplier is higher. I don't know why this is, but this is how it is. This used to be common knowledge before Telas training dummies were added (which you practically always did blood hits on, as they couldn't poof)

mrhultin:
> Based on Wiki and by going to official tibia you'll see Paladin skill distance faster than Knight level up melee, so making both equal is incorrect..
The reason for this, is that Blood hits/sparks (attacks where damage is fully absorbed by the monsters armor) are counted twice, not because their base multiplier is higher. I don't know why this is, but this is how it is. This used to be common knowledge before Telas training dummies were added (which you practically always did blood hits on, as they couldn't poof)

PR #2217, opened 2021-01-01
fix: stow crashing client

### LINKED ISSUE #2159: BUG: Server crash related to stash container
I am not using 100% clear base of this project but im pretty confident that my changes didnt affect that. Unfortunately I can't reproduce this crash and test on clear base.

Its very rare, it happened only twice during two months on test and main server played by 10-60 playes. We tried to reproduce this in many ways but nothing worked.

It happens when you try to stow an item, but I dont know if the way you do it even matters (drag or click stow on item)

[Heres crashlog](https://pastebin.com/Fkfx3g8Z).

gpedro:
current behaivor:
- go to depot
- create item 32648
- and drag to stash
if you open the stash, your client will crash.

---
we can just add an attribute `notstowable` to the item definition but will still be happening in the future with newer items. talking with Diego-OT, we told me about checking `wareId`, if has, you can stow, otherwise don't. Reading code around stash we actually have this on others part of the same system.

---
reported by : SergioS
credits to @Diego-OT

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -2170,7 +2170,6 @@
 	</item>
 	<item id="2148" article="a" name="gold coin">
 		<attribute key="loottype" value="gold" />
-		<attribute key="notstowable" value="1" />
 		<attribute key="weight" value="10" />
 	</item>
 	<item id="2149" article="a" name="small emerald">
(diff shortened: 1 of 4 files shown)


dex-89 (approved):
Works, but you need to remove the item from the stash

gpedro:
if the item is not stowable (RL) and you stow it you cannot remove from that because you will get a crash. to solve, only removing from table player_stash

dex-89:
> if the item is not stowable (RL) and you stow it you cannot remove from that because you will get a crash. to solve, only removing from table player_stash

Yes i know. I removed it and works

majestyotbr (approved):

PR #2236, opened 2021-01-04
fix: customize character crash caused by familiars

gpedro:
- right-click on your player > customize character
- you will get a crash

### DIFF
--- a/src/familiars.cpp
+++ b/src/familiars.cpp
@@ -30,6 +30,10 @@ bool Familiars::loadFromXml() {
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


Juanzin93:
tested, compiling got alot of warnings
image
and it didnt fix the issue, i go to my char in dawnport, change vocation to dawnport vocation, try to change outfits and crash with this error 
image

dex-89:
Dawnport: First vocation select -> Second vocation select (change) -> Customize character = crash

dex-89:
I checked Dawnport again today. Crash is only with **sorcerer**. Rest works ok

dex-89:
Fixed

PR #2240, opened 2021-01-04
Couple HoD fixes and the right gem names when using gems on the wrong shrines

### LINKED ISSUE #2078: FEATURE: heart_of_destruction
Lua Script Error: [Scripts Interface]
/home/global/data/scripts/creaturescripts/quests/heart_of_destruction/charger_spawn.lua:callback
luaAddEvent(). callback parameter should be a function.
stack traceback:
        [C]: in function 'addEvent'
        ...urescripts/quests/heart_of_destruction/charger_spawn.lua:23: in function <...urescripts/quests/heart_of_destruction/charger_spawn.lua:8>

oualid6496:
Fixes #2078


data/scripts/movements/quests/heart_of_destruction/vortex_crackler.lua
```diff
@@ -39,9 +39,7 @@ function vortexCrackler.onStepIn(creature, item, position, fromPosition)
 			local storePlayers, playerTile = {}
 			for i = 1, #positions1 do
 				playerTile = Tile(positions1[i]):getTopCreature()
-				if playerTile:isPlayer() then
-					storePlayers[#storePlayers + 1] = playerTile
-				end
+				storePlayers[#storePlayers + 1] = playerTile
```
gpedro: why removed the condition? and if is a summon or monster?

gpedro (changes requested):

data/scripts/movements/quests/heart_of_destruction/vortex_crackler.lua
oualid6496: removed it because it already checks if it's a player on line 32~36

	local player = creature:getPlayer()
	if not player then
		return true
	end

hyresu: But the only creature checked is the player who step's in, if there's a monster on the other positions they'll be counted as players on that loop with that removal, are you sure that's how it works?

oualid6496: @hyresu gonna test that out later today, but it shouldn't be counted since it's only being added when a creature is stepping in it right? Which shouldn't be happening anyway because of the first player check. Anyway im going to test this later today and revert it if it's indeed true

oualid6496: @hyresu @gpedro Tested it out and monsters/npcs aren't being registered as it should. So the monsters aren't counted when they're on the other positions.

marcosvf132: Some summons like the familiars can generate some problems in some cases, On my opinion i think it's better to have some extras safety checks if the playerTile really exist and if it is a player. You (and the old script) are using three external calls without any checks: positions2[i], Tile(positions2[i]) and Tile(positions2[i]):getTopCreature().

oualid6496: reverted it in new commit

data/scripts/creaturescripts/quests/heart_of_destruction/charger_spawn.lua
```diff
@@ -4,8 +4,8 @@ local function chargerSpawn(pos)
 	return true
 end
 
-local chargerSpawn = CreatureEvent("ChargerSpawn")
-function chargerSpawn.onDeath(creature)
+local chargerSpawns = CreatureEvent("ChargerSpawn")
```
dudantas: This change is not necessary, they are just variables
It also applies to the "cracklerTransforms"

dudantas (changes requested):

data/scripts/creaturescripts/quests/heart_of_destruction/charger_spawn.lua
oualid6496: Well actually, I was testing the quest out and it was bugging the script until I changed them

majestyotbr (approved):

dudantas (approved):

oualid6496 (commented):
@dudantas By reverting this the Cracklers stop changing into Depolarized Cracklers

data/scripts/creaturescripts/quests/heart_of_destruction/charger_spawn.lua
dudantas: > Well actually, I was testing the quest out and it was bugging the script until I changed them

I understood the error. I think there is something wrong with the code, and it is not changing the name of the global variable that this will solve, it just transfers the problem to another location.

majestyotbr:
So, add the fixes on #2292. Remember, when create a pull request you should use a branch based on develop and let develop untouched, to avoid conflicts.

PR #2241, opened 2021-01-05
Update schema.sql

ghost:
`isreward` varchar(255) DEFAULT NULL  added to the line 163


dudantas:
You need to explain ... What does this solve? Is it a bug? Is it an improvement? How do we test?

ghost:
Without this snippet, nothing is saved in the database

dudantas:
What do you mean, nothing is saved?
Isn't the database saving things?

Costallat:
@Madrugrelo I think that it's something that you customized on your server, there is no place on the code that saves that field 'isreward', that's why it doesn't exists on the schema.

PR #2265, opened 2021-01-12
Fix Letters

hyresu:
Fix for #2263

PR #2266, opened 2021-01-12
Fix -> the message of the letters are deleted

CarlosAlexanderGonzalezGarcia:
These changes are made by @hyresu
but it wasn't added, and you couldn't read what people wrote on the letters.
All credits are from @hyresu


hyresu:
Those changes were reverted as was inefficient, they were moved to mailbox.cpp and I made a new pull request to fix mailbox.cpp, if you can, test those again please #2265 
Thank you anyway for the intention!

PR #2277, opened 2021-01-15
add tables player charms

lmiguelm26:


dudantas:
Your contribution is very welcome, however, it is already added in the: #2278
Coincidentally, I added the migrations to the schema when you pulled it. Anyway, thank you very much!

PR #2252, opened 2021-01-08
Add wrap to house items

hyresu:
Please test it and give your feedback!

TODO:
- [ ] Add wrapable items to npcs instead of kit

PR #2281, opened 2021-01-15
New moveable beds

dex-89:
- [ ] wrap back function (#2252)
- [ ] deleting all beds on map
- [ ] `Unknown key value: sleepingPerson`
- [x] fix `Duplicate item with id: ...` ("id"="37376" -> "maleSleeper" value="37376", etc)
- [ ] found beds:id() in action file
- [ ] check `reserved items` in items.xml -> beds?

#1879 
```
Moveable Beds

You can now freely arrange any beds in your home! In order to make this happen, all beds will be removed from their
homes. You can buy new ones from furniture shopkeepers for 10 to 100 gp per bed. Every bed exists of two parts which
have to be purchased separately and will be placed separately. You can even put your bed on carpets! Each building
still has an individual bed limit however, so you can only put a certain amount of beds in them.
```
https://www.tibia.com/news/?subtopic=newsarchive&id=5272

### DIFF
--- a/data/items/items.xml
+++ b/data/items/items.xml
@@ -48940,7 +48940,6 @@
 	<item fromid="37292" toid="37295" article="a" name="ornate wall"/>
 	<item id="37296" article="a" name="ruined wall" />
 	<item fromid="37297" toid="37307" article="a" name="ornate wall"/>
-	<item fromid="37308" toid="37371" name="reserved sprite"/>
 	<item id="37372" toid="37373" article="a" name="luminescent crystal" />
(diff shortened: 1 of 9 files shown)


xThauma:
Nice to see my code. sleepingPerson param was my lazy idea to not let people use the bottom(legs) part of bed.

dex-89:
> Nice to see my code. sleepingPerson param was my lazy idea to not let people use the bottom(legs) part of bed.

If you have any idea how to fix it, I am waiting for suggestions

PR #2309, opened 2021-01-23
Fix loadLuaMapBook function

ElTeixe:
Fixes the loadLuaMapBook function not working properly. Is generating containers on the map when book item need to be generated at a map position and when the book item need to be in a container it isnt generated inside. Only works on existing already on map items.

- Fixed the issues, now generate new book items at map and inside containers, and still working on existing map items.
- Moved the loading console message inside the function.
- Merged the BookTable and DocumentTable. IMHO they are like the same, items that can be on map or inside a container that when are used they show a window with a readable text.
- Renamed function name and some changes on table variables.

NOTE: Although it said before there are 60 books and 2 documents, some books dont have position, so arent useable there are like 35, plus 2 documents. 37 now.

Now:

Fixed:

### DIFF
--- a/data/startup/others/functions.lua
+++ b/data/startup/others/functions.lua
@@ -89,20 +89,13 @@ end
 function loadLuaMapBookDocument(tablename)
 	-- Index 1: total valid, index 2: total loaded
 	local totals = {0, 0}
-	
-	-- Use ipairs to follow an numeric order (pairs order is unspecified)
-	-- So if we want create items in specific order, just have to define them on that order
-	-- (Although it only works for map items, container items are always added at first index. TO BE HANDLED)
 	for index, value in ipairs(tablename) do
 		-- Skip silently (some items dont have a know position yet defined, lets ignore them)
 		if not value.position then
 			goto skip
 		end
-		
 		totals[1] = totals[1] + 1
-		
 		local tile = Tile(value.position)
-		
 		-- Check if is a valid tile
 		if tile then
 			local item
@@ -110,24 +103,20 @@ function loadLuaMapBookDocument(tablename)
 			if value.containerId then
 				-- Try find the container on the map
 				local container = tile:getItemById(value.containerId)
-				
 				if not container then
 					print(string.format("> loadLuaMapBookDocument container not found! index: %d, containerId: %d", index, value.containerId))
 					goto skip
 				end
-				
 				-- Create the item inside the container
 				item = container:addItem(value.itemId, 1, INDEX_WHEREEVER, FLAG_NOLIMIT)
 			else
 				-- Try first find the item on the map (in some cases the item is already on the map)
 				item = tile:getItemById(value.itemId)
-				
 				-- Create the item at map position
 				if not item then
 					item = Game.createItem(value.itemId, 1, value.position)
 				end
 			end
-			
 			-- If the item exists, add the text
 			if item then
 				item:setAttribute(ITEM_ATTRIBUTE_TEXT, value.text)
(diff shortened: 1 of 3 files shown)


data/startup/others/functions.lua
```diff
@@ -86,26 +86,54 @@ function loadLuaMapSign(tablename)
...
-		local tile = Tile(value.itemPos)
-		local item
-		-- Checks if the position is valid
+function loadLuaMapBookDocument(tablename)
+	-- Index 1: total valid, index 2: total loaded
+	local totals = {0, 0}
+	for index, value in ipairs(tablename) do
+		-- Skip silently (some items dont have a know position yet defined, lets ignore them)
+		if not value.position then
+			goto skip
```
dex-89: Not better: (if position exist - read without skip and Github warnings: `luac: data/startup/others/functions.lua:95: '=' expected near 'skip'`)
```
if value.position then
  totals[1] = totals[1] + 1
  local tile = Tile(value.position)
  -- Check if is a valid tile
  if tile then
    local item
    -- Check if is a contained item
    if value.containerId then
      -- Try find the container on the map
      local container = tile:getItemById(value.containerId)
      if not container then
        print(string.format("> loadLuaMapBookDocument container not found! index: %d, containerId: %d", index, value.containerId))
      end
      -- Create the item inside the container
      item = container:addItem(value.itemId, 1, INDEX_WHEREEVER, FLAG_NOLIMIT)
      else
      -- Try first find the item on the map (in some cases the item is already on the map)
      item = tile:getItemById(value.itemId)
      -- Create the item at map position
      if not item then
        item = Game.createItem(value.itemId, 1, value.position)
      end
    end
    -- If the item exists, add the text
    if item then
      item:setAttribute(ITEM_ATTRIBUTE_TEXT, value.text)
      totals[2] = totals[2] + 1
    else
      print(string.format("> loadLuaMapBookDocument item not found or created! index: %d, itemId: %d", index, value.itemId))
    end
  else
    print(string.format("> loadLuaMapBookDocument tile not found! index: %d, position: x=%d y=%d z=%d", index, value.position.x, value.position.y, value.position.z))
  end
end
```

data/startup/others/functions.lua
```diff
@@ -86,26 +86,54 @@ function loadLuaMapSign(tablename)
...
-				item = Game.createItem(value.itemId, 1, value.itemPos)
-			end
-			if not item then
+			local item
+			-- Check if is a contained item
+			if value.containerId then
+				-- Try find the container on the map
+				local container = tile:getItemById(value.containerId)
+				if not container then
+					print(string.format("> loadLuaMapBookDocument container not found! index: %d, containerId: %d", index, value.containerId))
```
dex-89: Here you have the counter. This value cannot exceed 120
`data/startup/others/functions.lua:107:121: line is too long (127 > 120)`
we break the line where possible
image

dex-89 (changes requested):

data/startup/others/functions.lua
ElTeixe: I guessed that way to skip on the loop dont goes be loved lol.

- To the first case used on https://github.com/opentibiabr/otservbr-global/blob/5d589fdeb838af98aa4bf31192e71616d30754a4/data/startup/others/functions.lua#L95 yeah it may be added as a simple if as you show.
- But at second case used in https://github.com/opentibiabr/otservbr-global/blob/5d589fdeb838af98aa4bf31192e71616d30754a4/data/startup/others/functions.lua#L108 (if the container dont exist) instead to jump to next item at end of the loop, will continue to process until the line https://github.com/opentibiabr/otservbr-global/blob/5d589fdeb838af98aa4bf31192e71616d30754a4/data/startup/others/functions.lua#L131 and will be useless since if there is no container will be no item. 

I dont know if I explained myself properly.

data/startup/others/functions.lua
ElTeixe: About that totally right, that line lengths gonna to kill me.. lol I dont really have the habit to use that lower ones. I'll fix them.

dex-89: It's not most important.

data/startup/others/functions.lua
ElTeixe: BUMP! I edited the reply, I'll make a newer one next time or may be confusing.

dex-89: I understand. I think that @dudantas should comment

dex-89 (approved):

majestyotbr (approved):

PR #2320, opened 2021-01-27
Fix: Bestiary 100% CPU and search method.

### LINKED ISSUE #2319: BUG: Server freeze (100% CPU) when typing space into bestiary monster search
Basically what title says. It happened twice on my test server but now I have found a way to reproduce this issue. I don't know if typing space is the only way, other characters I tried were fine. 

How to reproduce: Open cyclopedia -> bestiary window, type space into search window, and then click search.
[Video showing the issue](https://gfycat.com/velvetybetterblueandgoldmackaw)

Perf top command showing 100% of cpu usage (server does not crash, just freezes until manually killed)
ApplicationFrameHost_ep3zP6ErPG

Theres also another kinda related issue with bestiary search, it always shows only one creature instead of each one containing the word you typed. For example if you search for orc, it will show only one of orcs instead of each known one.

Related to both issues: if you type space in real tibia it will show each of known creature that contain space in their name.

marcosvf132:
Resolves #2319
Fix the error when searching for a big number of monsters on the 'bestiary' window and changing the search method to find all monsters according with the search words.

### DIFF
--- a/src/protocolgame.cpp
+++ b/src/protocolgame.cpp
@@ -1899,28 +1899,17 @@ void ProtocolGame::parseBestiarysendCreatures(NetworkMessage &msg)
 	std::string text = "";
 	uint8_t search = msg.getByte();
 
-	if (search == 1)
-	{
+	if (search == 1) {
 		uint16_t monsterAmount = msg.get<uint16_t>();
 		std::map<uint16_t, std::string> mtype_list = g_game.getBestiaryList();
-		uint16_t raceid = msg.get<uint16_t>();
-		MonsterType *mtype = g_monsters.getMonsterTypeByRaceId(raceid);
-		if (!mtype)
-		{
-			return;
-		}
-		text = mtype->info.bestiaryClass;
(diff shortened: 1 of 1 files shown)


gerotib:
@marcosvf132 Tested and it works fine now, only thing that differ from rl is that it shows "unknown" searched creatures, and on rl it wouldn't show them at all, but well thats not critical :D

gerotib (approved):

marcosvf132:
> @marcosvf132 Tested and it works fine now, only thing that differ from rl is that it shows "unknown" searched creatures, and on rl it wouldn't show them at all, but well thats not critical :D

Try it now, i added a filter to show only monsters that the player already unlocked.

gerotib (approved):
Works as it should now

dex-89 (approved):

ElTeixe (approved):
Checked. Fixes the issue and the filter of killed creatures is working. Good work!

majestyotbr (approved):

PR #2322, opened 2021-01-27
feat: add webhook notifications for server start, shutdown, server save and raids

wizardlink:
This PR adds webhook notifications to Discord.

The URL can be configured through the config: discordWebhookURL

This also adds the libraries libcurl and libjsoncpp as dependencies.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,8 @@ install:
   - cmd : vcpkg install boost-system:x64-windows
   - cmd : vcpkg install boost-variant:x64-windows
   - cmd : vcpkg install cryptopp:x64-windows
+  - cmd : vcpkg install curl:x64-windows
+  - cmd : vcpkg install jsoncpp:x64-windows
   - cmd : vcpkg install libmariadb:x64-windows
   - cmd : vcpkg install luajit:x64-windows
   - cmd : vcpkg install mpir:x64-windows
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -36,7 +36,7 @@ jobs:
             triplet: x64-osx
             packages: >
               boost-asio boost-iostreams boost-system boost-filesystem boost-variant boost-lockfree
-              lua libmariadb pugixml cryptopp
+              lua libmariadb pugixml cryptopp curl jsoncpp
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/build-mac.yml
+++ b/.github/workflows/build-mac.yml
@@ -36,7 +36,7 @@ jobs:
             triplet: x64-osx
             packages: >
               boost-asio boost-iostreams boost-system boost-filesystem boost-variant boost-lockfree
-              lua libmariadb pugixml cryptopp curl jsoncpp
+              curl jsoncpp lua libmariadb pugixml cryptopp
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -33,10 +33,10 @@ jobs:
     - name: Install Dependencies
       run: >
         sudo apt-get update && sudo apt-get install cmake build-essential ca-certificates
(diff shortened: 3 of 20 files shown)


Costallat (approved):

wizardlink:
Since @Costallat has approved, I suppose we can move on even though we have a check failing? Will wait for your input @MajestyOTBR.

dex-89:
> Since @Costallat has approved, I suppose we can move on even though we have a check failing? Will wait for your input @MajestyOTBR.

discordWebhookURL = invite discord url?

wizardlink:
> > Since @Costallat has approved, I suppose we can move on even though we have a check failing? Will wait for your input @MajestyOTBR.
> 
> discordWebhookURL = invite discord url?

_Responding here too for the sake of keeping a record in the repository._

That configuration is to use a webhook from a Discord server, you can find that here:

image
image

dex-89 (approved):

Tazer89:
Tested and works like charm. Should have luascript to it so it easy to add custom notifications :) Anyway awesome work!

wizardlink:
> Tested and works like charm. Should have luascript to it so it easy to add custom notifications :) Anyway awesome work!

Haven't thought about that but it sounds like a nice idea. Since then you can add things like login messages more easily and whatnot. I might consider adding that once I am satisfied with my spell system.

dex-89 (approved):

wizardlink:
@dex-89, @MajestyOTBR. That's my last commit for this pull-request, whether it moves forward is up to you guys.

For the issue we encountered for the custom maps not loading: Something has fixed it recently, both locally and on my server, everything is being loaded correctly.

majestyotbr (approved):

majestyotbr (approved):

majestyotbr:
Thanks for your contribution!

wizardlink:
> @wizardlink there is a **memory leak** and **crash**, just test it with some talkactions (do it 10x per example and check you ram)
> @MajestyOTBR
> 
> PS. you can use /teste 10x (fast) with Webhook.send("Teste", "testing", WEBHOOK_COLOR_WARNING)

This is a _limitation_ in a sense with the curl library, the fix as of now would be to do global cleanups instead of cleaning after every request but it doesn't makes sense to have that if, for example, there are multiple requests on-going. I'd say that you can use a global clean if you'd like, but I'd recommend just avoiding making unnecessary requests, at least for now.

In the long run, it would be beneficial to create a file for suppressions and check for curl.

wizardlink:
> You mean curl_global_cleanup() ? I was tested and same problem =/
> 
> Edit: 
> Each /test
> appear on my screen:
> [New Thread 0x7ffff0a20700 (LWP 23741)]
> [Thread 0x7ffff0a20700 (LWP 23741) exited]

Interesting, mind giving me the basic specs of the machine (including OS)?

Also in regards to what your friend said, there is no such thing. It is just sending a HTTP request to Discord.

wizardlink:
> Distributor ID: Ubuntu
> Description: Ubuntu 16.04.7 LTS
> Release: 16.04
> Codename: xenial
> Processor: AMD Ryzen 5 3600X - 6 c / 12 t - 3.8 GHz / 4.4 GHz
> Memory: 32GB
> 
> One question, is possible send image to discord?
> Thanks!

Sorry for the delay, had a full day. Yes, you can send an image to Discord, just not through this implementation. You can check out this for more info: https://discord.com/developers/docs/resources/channel#embed-object

In regards to the overall problem, I should come up with a solution until Wednesday. I already found the issue, just figuring out the best way to tackle it. :)

-Edit-

FYI: A segment fault is occurring because the project compiles to 32bit, so, it only uses a maximum of 4GB. Which brings me to mention... @MajestyOTBR, you should talk with your development team about it. I don't think it's beneficial to maintain it at this day and age at 32 bits.

-Edit 2-

I was way too abroad with my earlier statement, specifically, this error is occurring when the `write_data` function is being called. You can do a quick fix by just removing it's reference and the method itself. However, I am looking into how that function is using so much memory that the application can't allocate more.

PR #2333, opened 2021-01-30
Fix vocation adjustments + Weapon formula correction

hyresu:
This will fix the following things:
- Burst Arrow not using elemental damage

And will add a feature that was missing that is the formula correction for weapons as Global stated:
"The minimum damage of single target, non-critical physical range and melee attacks has been increased to avoid very low hits. While their maximum damage has not been changed so that occasional high damage spikes are still possible, its probabilities was lowered to ensure that the average damage output remains the same."

The formula isn't that acurrate, it must be optimized to become the same avg hit as the old, though they are similar it must be more precise.

TODO :
- Don't use the formula correction on crit

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -23,6 +23,11 @@ orangeSkullDuration = 7
 onlyInvitedCanMoveHouseItems = true
 cleanProtectionZones = false
 
+-- Formula Correction
+formulaCorrection = true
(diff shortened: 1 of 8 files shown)


Bipbop69 (approved):

majestyotbr:
PR closed due to inactivity.

PR #2295, opened 2021-01-18
Map fixes

dex-89:
- [X] Fixed exit teleport position (Cobra bastion: Scarlet) to 33395, 32670, 6
- [X] Fixed depot's id
- [X] Fixed city name [citizen teleport Liberty bay]
- [X] Fixed issues: #2196, #2303
- [X] Checked and fixed `Thieves guild quest` #1663 
- [X] Fix return Astral Shaper Boss Dungeon #2315 
- [X] Added PZ on Dawnport entrance
- [X] Fixed cure poison tiles action ids at Dawnport outpost entrances by @ElTeixe 
- [X] Added npc `Stricken soul` to map #2229
- [X] Added `Safe room` energetic library #2057 
- [X] Fixed doors (open/closed) in Liberty bay by @wizardlink 

**Fixed Rookgaard quests:** [#2347]
- [X] Fixed `Bear Room Quest` `To retrieve the key, use a pick on the mud floor square directly south of the big table.`
- [X] Fixed `Captain Iglues Treasure Quest` (left reward chest was not working)
- [X] Fixed `Honey flower quest` (the reward of a honey flower, not a chest)
- [X] Fixed closed door to minotaur mage (Rookgaard) now player cant open it
- [X] Added missing `dead human` [`Pick quest`] (way to `katana quest`)
- [X] Added and fixed missing items at Rookgaard (The Rookie guard quest) by @ElTeixe 
- [X] Fixed `hole -> trap` by @ElTeixe 

**Added/fixed in global update:**

Source: https://www.tibia.com/news/?subtopic=newsarchive&id=5970&fbegind=10&fbeginm=1&fbeginy=2021&fendd=9&fendm=2&fendy=2021&flist=11111111
- [X] added new item to black bert and fix item's name
- [X] added northward balcony (Fortune Wing 4 Yalahar)

**Special thanks: @ElTeixe for help**

PR #2370, opened 2021-02-07
fix(Movements): liberty bay portal of citizenship

wizardlink:
This resolves the problem in _Liberty Bay City_ where if you enter the portal of citizenship, you become a citizen of Darashia instead.


dex-89:
Added in #2295

PR #2325, opened 2021-01-28
feat: feature/decouple everything and sync with SaiyansKing optimized forgottenserver

dudantas:
All the global variables g_* were decoupled to true singletons, reducing the dependencies between the objects and simplifying the amount of calls and instances. They used to be partial singletons mixed with global variables, that way we have one, and only one instance that is accessed via the getInstance methods, that are now the g_* functions.

g_luaEnvironment is a little bit more complex than the others. Due to that, it's the only one that remained in the old format.

* Decouple: actions
* Decouple: ban
* Decouple: chat
* Decouple: config
* Decouple: configmanager
* Decouple: creatureevent
* Decouple: database
* Decouple: databasetasks
* Decouple: dispatcher (tasks)
* Decouple: events
* Decouple: game
* Decouple: globalevent
* Decouple: imbuements
* Decouple: luascript
* Decouple: modules
* Decouple: monsters
* Decouple: movements
* Decouple: rsa
* Decouple: scripts
* Decouple: spells
* Decouple: talkactions
* Decouple: vocations
* Decouple: weapons

* Instantiate g_luaEnvironment to enable tests

* Add:
a few more player safeness
string_extend.h
robin_hood.h
scripts.cpp/h

* Fix summons/parent tile segmentation fault for creatures

* Remove:
scriptmanager
scheduler
xtea

* Syncronized changes from @SaiyansKing optimized forgotten server
https://github.com/SaiyansKing/optimized_forgottenserver

* Credits from the decouple for the @lgrossi
https://github.com/CanaryProject/canary-server/pull/19

PR #2372, opened 2021-02-07
fix: correct editorconfig

wizardlink:
Adjusts the [editorconfig](https://editorconfig.org/) to reflect the actual indentation style of the project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -22,9 +22,8 @@ indent_size = 2
 indent_style = space
 indent_size = 2
 
-# Google Style
 [*.{cpp,hpp,h}]
-indent_style = space
+indent_style = tab
 indent_size = 2
 indent_brace_style = K&R
 spaces_around_brackets = none


dex-89 (approved):

.editorconfig
```diff
@@ -22,9 +22,8 @@ indent_size = 2
 indent_style = space
 indent_size = 2
 
-# Google Style
 [*.{cpp,hpp,h}]
-indent_style = space
+indent_style = tab
```
dudantas, line 26: The correct is space (although almost all indentation is as a tab), because the tab bugs the indentation on github

dudantas (changes requested):

.editorconfig
wizardlink: Since when?

dudantas: Since when, what?

wizardlink: Since when does it _breaks on GitHub_?

You might be mistaking the showcasing of mixed indentation, if you have a line with space indentation and another with a tab, their width will be completely different.

However, that has never been an issue and you have the same behaviour in GitLab and Azure. It is not a GitHub problem.

The only _fix_ for that is having a single indentation, not mixed.

dudantas: I didn't say it breaks, I said it bugs (visually).

dudantas: I can give you an example ... Open my PR #2325, go to game.cpp, open the diff, search for:
`if (player->getPathTo(item->getPosition(), listDir, 0, 1, true, false)) {
			g_dispatcher().addTask(std::bind(&Game::playerAutoWalk,
											 this, player->getID(), listDir));`
											 
This is with TAB, check that when using space, the indentation is correct
Tthe "this" should be just below the "& Game :: playerAutoWalk,"

wizardlink: And that falls exactly on what I said, mixed indentation. If the indentation of both lines were space, or tab, then this _issue_ would not occur.

dudantas: No. All indentation is done with the tab. The problem is that 1 tab is identified by only 1 character (when in fact there are 4) and github does not identify it.
https://softwareengineering.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e

wizardlink: I see what you mean, but, that's still not a problem. As well as that answer being... Partly incorrect in this context. I took a look at what you meant with that file now, since, you didn't provide a direct link to the line and I was eating lunch.

By default in any source control visualizer, you have a tab being represented by eight spaces. Due to legacy, it has been kept like that, it would take an unnecessary processing to determine the `tabstop` of your project. That is why GitHub (and other source control visualizers) have query options. In GitHub you can pass `?ts=x` to determine the tabstop you are seeing, but, you can also have a [`editorconfig`](https://editorconfig.org/) on the root of the project and GitHub will use that to determine the `tabstop`.

With that out of the way, here is where things get complicated: This project has the wrong configuration for it's `cpp` indentation. When you are telling GitHub that `cpp` files use spaces for indentation, it can't guess what's the `tabstop`.

An example of this in the work: https://github.com/Hydractify/otservbr-global/blob/develop/src/game.cpp

You will notice that in my repository, we have the correct configuration, thus, you can see the actual indentation of the file.

In summary: GitHub looks at your file and guesses the `tabstop` since the [editorconfig](https://editorconfig.org/) is incorrect. It also will treat it as a file with mixed indentation since, to GitHub, the correct indentation is 2 spaces.

Costallat: We use editorconfig not to configure Github but to configure our local environment(BTW: I didn't know that Github also read that file), if you change that to tab, my editor will start to use tab in place of space, as we are trying to follow the google style, we need to use [space](https://google.github.io/styleguide/cppguide.html#Spaces_vs._Tabs) or we are going to start mixing several styles again, that why it is configured to be spaces in place of tab

wizardlink: I agree with you, it is meant as a local fix. GitHub isn't a problem like it was mentioned previously, just a nice to have.

However, I don't believe enforcing a style that isn't in pair with the project is a good idea. If you believe that space indentation is the way to go, I will gladly accept it, after all, I created this PR to hopefully achieve consistency in this project. That said, if you want to use spaces, then you must first change the project adequatily and not keep the mixed indentation as-is.

You can't try to create consistency by writing inconsistencies. A reminder that this project doesn't has a workflow that checks for the indentation, so the `.editorconfig` is for those contributors who are aware of it's existence.

wizardlink: I usually wait for a reply before rebutting, but, here is just hard proof of how `space` is actually bringing more inconsistency than consistency:
image

Even if you divide the tabs by two (considering they are a single character, while the 2 space indentation, are two) it still is 200.000 more occurrences. `space` indentation makes up for 11% of the code, if you don't divide, that is less than 6%.

Edit: I apologise, it's 12% of the code when you divide, and 6% if you don't. I didn't account for the start of line anchor and catching all tabs in a single line. Which that removes 639 matches.

dudantas:
I'm starting major changes in the source, gradually changing the indentation to SPACE instead of TAB, until then, we will continue using TAB as the default.

PR #2379, opened 2021-02-08
Fix bath tub

hyresu:
The bath tub actually uses 3 ID
29312 not filled
29313 filled but movable
29336 filled but not movable

The behavior expected is that you can move the 29313, it was locked via items.xml when on otb it allowed to be moved.

Does anyone have a bath tub in global for testing? I think that 29313 is actually wrapable.


ElTeixe (approved):

data/scripts/movements/others/bath_tub.lua
```diff
@@ -6,19 +6,12 @@ function bathtubEnter.onStepIn(creature, item, position, fromPosition)
...
-        return true
-      end
-    end
-
     local condition = Condition(CONDITION_OUTFIT)
     condition:setOutfit({lookTypeEx = playerBathTub})
     condition:setTicks(-1)
-
+	
+	position:sendMagicEffect(CONST_ME_WATERSPLASH)
```
dudantas, line 13: Wrong indent?

ElTeixe (approved):
Tested. Previous bug fixed, correct descriptions and correct way to fill it with water.

data/scripts/movements/others/bath_tub.lua
dex-89: Fixed

dex-89 (approved):

majestyotbr (approved):

PR #2403, opened 2021-02-13
Shatered island quest bug

jollyrogers91:
Added dot between storage and shatered isles quest to make storage work.


henriquesameshima (changes requested):
You have duplicated the lines and actions within the file. Note that just below you have the same lines.

henriquesameshima:
@jollyrogers91 added corrections in PR #2407 correctly, thanks.

PR #2453, opened 2021-02-25
Removed hardcoding monsters list.

henriquesameshima:
- Now the monster list is created / loaded when starting the server using the monsters added in the map respawns, this means that whenever a new monster is added to the map, it will be included in the prey monster list.
- Added boss checking to prevent it from being included in the prey.


ElTeixe:
I thinked about it the first time you improved the prey system, but as you seem dont understand me I stop made suggestions. Lets see if I can explain myself this time... 

What about instead of that, use the current existing function to get the bestiary list of monsters with: `Game.getBestiaryList()` it returns all monsters which have bestiary (bosses are excluded due they dont have bestiary) so to a monster get included just need that the monster have its bestiary data.

As the list returns all monsters with bestiary will return also the monsters that give 0 exp, but we can exclude them before start use the list instead check it at runtime while picking a new random monster.

As we are accesing monster bestiary data, we can store on the list the stars too and avoid create a MonsterType instance to check its difficulty later.

---

Then we have 2 new functions at (`data/modules/scripts/prey_system/prey.lua`) one to initialize the `Prey.MonsterList` and another function to return a valid list of prey creatures:

```
Prey.MonsterList = nil

-- Returns a valid prey list of monsters containing its name and its difficulty tier
function Prey.getPreyList()
	local preyList = {}
	local bestiaryList = Game.getBestiaryList()
	
	for i = 1, #bestiaryList do
		local monster = MonsterType(bestiaryList[i])
		-- Check monster exists and it gives exp
		if monster and monster:experience() > 0 then
			preyList[#preyList + 1] = {name = monster:name(), difficulty = monster:BestiaryStars()}
		end
	end
	
	return preyList
end

-- Initialize the prey list to future uses
function Prey.init()
	if Prey.MonsterList == nil then
		Prey.MonsterList = Prey.getPreyList()
	end
end
```

We can initialize the `Prey.MonsterList` at startup at (`data/scripts/globalevents/others/startup.lua`) like this:

```
-- Prey System Init
Prey.init()
```

Then we can use it on the current function `Player.createMonsterList()` (I added too some improvements in the checks):

```
function Player.createMonsterList(self)
	-- Do not allow repeated monsters
	local repeatedList = {}
	for slot = CONST_PREY_SLOT_FIRST, CONST_PREY_SLOT_THIRD do
		if (self:getPreyCurrentMonster(slot) ~= '') then
			repeatedList[#repeatedList + 1] = self:getPreyCurrentMonster(slot)
		end
		if (self:getPreyMonsterList(slot) ~= '') then
			local currentList = self:getPreyMonsterList(slot):split(";")
			for i = 1, #currentList do
				repeatedList[#repeatedList + 1] = currentList[i]
			end
		end
	end
	-- Generating monsterList
	local monsters = {}
	local counters = {
		[Prey.Difficulty.TRIVIAL] = 0,
		[Prey.Difficulty.EASY] = 0,
		[Prey.Difficulty.MEDIUM] = 0,
		[Prey.Difficulty.HARD] = 0,
		[Prey.Difficulty.CHALLEGING] = 0
	}
	while (#monsters ~= 9) do
		local monster = Prey.MonsterList[math.random(#Prey.MonsterList)]
		-- Verify that monster is not repeated and check tier max amount
		if not table.contains(repeatedList, monster.name) and counters[monster.difficulty] < getMaxMonster(self, monster.difficulty) then
			-- Add it to monsters list
			monsters[#monsters + 1] = monster.name
			counters[monster.difficulty] = counters[monster.difficulty] + 1
		end
		-- Add it to repeated list
		repeatedList[#repeatedList + 1] = monster.name
	end
	return table.concat(monsters, ";")
end
```

And finally we can delete the function `getMonsterDifficulty()` that isnt needed anymore since we already have the monster difficulty on the `Prey.MonsterList`.

I did a branch with that changes and seems to work properly: https://github.com/opentibiabr/otservbr-global/compare/develop...ElTeixe:bestiary-fix?expand=1

The only exception is that if the monster dont have defined its bestiary data will not be included on the list like at this moment happens with the Terrified Elephant (I tried find its id.. but wasnt able).

I hope being understood, and if not you can ask me about.

henriquesameshima:
WOW, much better changes than mine, you could have indicated that earlier. I will close my PR.

PR #2474, opened 2021-03-11
Fix an issue on reload spells

UReddington:
Just a fix of an issue on reload about spells that was reloading monsters instead spells.

### DIFF
--- a/src/signals.cpp
+++ b/src/signals.cpp
@@ -153,7 +153,7 @@ void Signals::sighupHandler()
 	g_game.raids.startup();
 	std::cout << "Reloaded raids." << std::endl;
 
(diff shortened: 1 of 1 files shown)


phacUFPE:
I think that the spells reload are fine, did you mean about the /reload spells ?

UReddington:
I'm sure that's related to the reload all function, but it's still being a bug that the monsters get reloaded instead spells (not even sure if even monsters should be reloaded)

phacUFPE (changes requested):
Fix compilation errors

UReddington:
Listo

dudantas (approved):

PR #2486, opened 2021-03-19
Fix: quest chest system

UReddington:
This chest contains a small fix for the quest system, it removes the multiple messages on window and also removes a bug of invalid keys getting generated.

### DIFF
--- a/data/scripts/actions/system/quest_reward_common.lua
+++ b/data/scripts/actions/system/quest_reward_common.lua
@@ -5,18 +5,19 @@
 local AttributeTable = {
 	[6013] = {
 		text = [[
-Hardek *
-Bozo *
-Sam ****
-Oswald
-Partos ***
-Quentin *
-Tark ***
-Harsky ***
-Stutch *
-Ferumbras *
-Frodo **
-Noodles ****]]
+			Hardek *
+			Bozo *
+			Sam ****
+			Oswald
+			Partos ***
+			Quentin *
+			Tark ***
+			Harsky ***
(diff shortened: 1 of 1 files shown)


joao-blip (approved):

adrianomengue:
@UReddington could you explain me how to reproduce the bug? So I can test your PR.

dudantas (commented):
I did not understand. What was the correction?

data/scripts/actions/system/quest_reward_common.lua
```diff
@@ -5,18 +5,19 @@
...
-Oswald
-Partos ***
-Quentin *
-Tark ***
-Harsky ***
-Stutch *
-Ferumbras *
-Frodo **
-Noodles ****]]
+			Hardek *
```
dudantas: You need to get back to how it was before. If you are not going to show this space that you gave, there in the book.

UReddington:
> 
> 
> I did not understand. What was the correction?

When you open a chest with 2 a key it seems like you don't get it with the valid actionid (required to open the doors), but testing it recently i found that the fix isn't enough, so i wouldn't recommend to push this for now.

LukeAtMyNose (commented):
In my test with the copper key 3801, the key now does not appear in the bag

image

UReddington:
> 
> 
> In my test with the copper key 3801, the key now does not appear in the bag
> 
> image

this commit is incomplete, im going to close it until we have a better solution for this problem

PR #2499, opened 2021-03-30
Full rework in the npcs method/class

dudantas:
- [x] Reworked the storage of npcs to the world/otservbr-npc.xml file
- [x] Reworked the class of npcs in the source, with behavior similar to that of monsters
- [x] Created NpcType
- [x] Reworked the onSay module
- [x] Recreated all npcs with the method for revscriptsys (thanks @gpedro)
- [x] Reworked the basic behavior of npcs
- [ ] Reworked the lib of the npcs
- [ ] Reworked the npcs scripts
- [ ] Reworked the onTrade module

Known bugs
- Npcs don't walk

This pull request is a work in progress (wip).

My intention is to completely recreate the class of npcs (creating NpcType), based on monsters. This will allow numerous customizations to the npcs (from npc that attack player, use spell, drop loot, to callbacks made in a simple and organized way, without as much work as it was before).

It is a job that will take a while to get ready, but, at the request of project members, I release it in parts, so that they can follow up, help with tests and give tips, doing a team work.

Keep in mind that we will also return to the old way of storing npcs by XML (I mean, the XML file in the world folder), but in a more organized way, as you can see in the pull request https://github.com/opentibiabr/rme/pull/3/files in the otservbr rme fork.

With that, we hope to be able to make life easier for those who use the editor, completely decoupling the npcs from the monsters, both in the editor and on the server.
At the present time, all of the old lib/modules from the npcs have been removed, so that I can work EVERYTHING from scratch. The basic behavior of npcs is working, from being created on the map, to its main functions (also hi/bye, only).

Any bugs found, should be reported here, in the comments.


Bipbop69:
what reason for this change? was causing a problem?

dudantas:
> what reason for this change? was causing a problem?

this type of table in Lua was not good ... it was interesting the npcs on the map, because you could see them there. this change allows it to be changed by both the map and the xml, the Lua only changes by the Lua table itself ... so it's a improvement...

data/lib/npc/npc.lua
```diff
@@ -11,46 +11,48 @@ end
...
-	if msgContains(message, "hi") and self:isInTalkRange(creature:getPosition()) and not self:getFocus() then
+function Npc:greet(message, creature, text)
+	if msgContains(message, "hi") and self:isInTalkRange(creature:getPosition()) and not self:isInteractingWithPlayer(creature) then
+		self:addPlayerInteraction(creature:getId())
 		if text then
-			self:sendMessage(text)
+			self:sendMessage(creature, text)
 		else
-			self:sendMessage("Hello, ".. creature:getName() ..", what you need?")
+			self:sendMessage(creature, "Hello, ".. creature:getName() ..", what you need?")
```
lgrossi, line 39: "what do you need?"

data/npc/a_bearded_woman.lua
```diff
@@ -48,33 +48,35 @@ npcType.onMove = function(npc, creature, fromPosition, toPosition)
...
-	if creature:getPlayer() then
-		if npc:greetMessage(message, creature, "GET ME OUT OF HERE! NOW!") then
+	local player = creature:getPlayer()
+	if player then	
+		if npc:greet(message, player, "GET ME OUT OF HERE! NOW!") then
 			return true
-		elseif npc:farewellMessage(message, creature) then
+		elseif npc:unGreet(message, player) then
 			return true
 		end
```
lgrossi:
```suggestion
		if npc:startConversation(message, creature, "GET ME OUT OF HERE! NOW!") then
		    return true
		end
```

data/npc/a_bearded_woman.lua
```diff
@@ -88,26 +90,28 @@ npcType.onSay = function(npc, creature, type, message)
...
-				npc:talk(value)
-				return true
+		if npc:isInteractingWithPlayer(player) then
+			for index, value in pairs(messages) do
+				if msgContains(message, index) then
+					npc:talk(player, value)
+					return true
+				end
 			end
 		end
```
lgrossi:
```suggestion
for index, value in pairs(messages) do
	if msgContains(message, index) then
	   npc:talk(player, value)
	   return true
   end
end
```

dudantas:
Now this is here: https://github.com/opentibiabr/canary

PR #2519, opened 2021-04-08
Prevent gold cloning

lBaah:
High ranges allow players to duplicate money when repeatedly betting on a high or low and doubling the bet when they lose.

### DIFF
--- a/data/npc/scripts/custom/cassino.lua
+++ b/data/npc/scripts/custom/cassino.lua
@@ -1,7 +1,7 @@
 local config = {
 	bet = {
-		min = 10000, -- gold coins // 30k
-		max = 10000000000, 
+		min = 100000, -- gold coins // 30k
+		max = 1000000, 
 		win = 180, -- 170% high/low
 		winNum = 500, -- 300% numbers
 	},


oualid6496:
M8 that isn't gold cloning at all, it's just how it works. Ofc you can keep doubling it but you will only make the chance higher to lose it all or win minimal profit.

lBaah:
First of all, I'm not your "m8".
I managed to get infinite money on few servers, so it's an exploit.
You will NEVER lose gold betting lower as possible and doubling every time you lose.

lgrossi:
> First of all, I'm not your "m8".
> I managed to get infinite money on few servers, so it's an exploit.
> You will NEVER lose gold betting lower as possible and doubling every time you lose.

M8, please be kind with who is trying to help.

That's not a bug, that's bad configuration. When you say "cloning" money it sounds like a bug, someone duplicating money, exploiting a problem. Thats no true, what happens is a bad config that makes things too easy. 
If we wanna fix that, its fine. Just make sure you proper explain it and also add reasons for the values used. Random values without proper argumentation are not helpful.

lBaah:
I'm not yout m8 either.
People are downloading and running this as it is, then let that happen if you will. :)

adrianomengue:
I went to try your method. I gave 1kk to my tester player. 
10k bet, after lose 20k bet, 40k, etc....

After some minutes, going up and down, I lost everything.

oualid6496:
Chill, didn't want to trigger you. This might be interesting to read https://en.m.wikipedia.org/wiki/Martingale_(betting_system)

cys9211:
Maybe someone has a script similar to how gunzodus has for npc Cassino

oualid6496:
@cys9211 not familiar with their cassino how is it different then this one?

adrianomengue:
> Chill, didn't want to trigger you. This might be interesting to read https://en.m.wikipedia.org/wiki/Martingale_(betting_system)

It's a very good way to breake your bank ^.^

lgrossi:
This PR makes no sense. Changing the bet min/max won't fix a reward issue. It should either review the reward rates or the bet modes, but not how much you should bet. You still have a 50% bet that has 1.8 profit return. 
One approach would be splitting the rates (the number bet can have a higher profit return while the 50% ones, odd, even, high, low, should have something lower, 40% or so).

PR #2520, opened 2021-04-08
Conversion of npcs to revscriptsys and loading spawns to XML (only source code)

dudantas:


adrianomengue:
Will hirelings be affected by this PR? I mean, it works as a NPC.

dudantas:
> Will hirelings be affected by this PR? I mean, it works as a NPC.

probably... when I finish I see what the hirelings will look like

dudantas:
Now is here: https://github.com/opentibiabr/canary

PR #2567, opened 2021-05-03
Fix restart prey after kick from beds by owner in house

dex-89:
- [x] Fixed restart prey after kick from beds by owner in house

### DIFF
--- a/src/creatures/players/player.h
+++ b/src/creatures/players/player.h
@@ -132,7 +132,7 @@ struct Kill {
 
 using MuteCountMap = std::map<uint32_t, uint32_t>;
 
-static constexpr int32_t PLAYER_MAX_SPEED = 15000;
+static constexpr int32_t PLAYER_MAX_SPEED = 4500;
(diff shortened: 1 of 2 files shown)


ghost (approved):

src/creatures/players/player.h
```diff
@@ -132,7 +132,7 @@ struct Kill {
 
 using MuteCountMap = std::map<uint32_t, uint32_t>;
 
-static constexpr int32_t PLAYER_MAX_SPEED = 4500;
+static constexpr int32_t PLAYER_MAX_SPEED = 15000;
```
lgrossi: Im not sure if this is the correct approach. We shouldn't raise the bars for normal players. This would allow level, spells and items to stack more, not only game master.

dex-89: You're right. I didn't think about it. Thank you for your attention.

majestyotbr (approved):

dudantas (approved):

PR #2580, opened 2021-05-12
Fix 'summonable', 'convinceable' and 'pushable' flags of Hero Fortress monsters

Aerwix:
Creatures Altered:
- Vicious Squire
- Vile Grandmaster
- Renegade Knight
- Arachnophobica

I modified this monsters because they can be:
- Summoned
- Convinced
- Killed when other monsters 'push' them.
- Can't push items So, you can put parcels and do free level)


oualid6496:
Are you sure they're summonable? That doesn't seem right...

lgrossi (approved):

lgrossi:
Can you update the description please, its a bit confused. What you mean is that they shouldn't be but currently they are, right?

dudantas (approved):

PR #2615, opened 2021-06-07
Fix imbuing deffs count to character combat stats

### LINKED ISSUE #2603: BUG: Elemental protection imbuiments not working
We are able to enchant our gear with Elemental Protection Imbuiment, like Dragon Hide (Fire resist), but the elemental % damage reduction is not applied to our character.

Any idea how to fix?

ArturKnopik:
fix #2603


ArturKnopik:
fix i variable

majestyotbr (approved):

src/server/network/protocol/protocolgame.cpp
```diff
@@ -3000,43 +3000,73 @@ void ProtocolGame::sendCyclopediaCharacterCombatStats()
...
+				if (slots)
+				{
+					for (uint8_t imbuSlot = 0; imbuSlot < slots; imbuSlot++) {
+						uint32_t info = item->getImbuement(imbuSlot);
+						if (info >> 8) {
+							Imbuement* ib = g_imbuements->getImbuement(info & 0xFF);
+							absorbs[i] += ib->absorbPercent[i];
+						}
+					}
+				}
```
lgrossi: You can simplify this whole part considerably..

```
  for (int32_t slot = CONST_SLOT_FIRST; slot <= CONST_SLOT_LAST; ++slot) {
		Item* item = player->getInventoryItem(static_cast<slots_t>(slot));

		if (!item) {
			continue;
		}

		const ItemType& it = Item::items[item->getID()];
    
		for (uint16_t i = 0; i < COMBAT_COUNT; ++i) {
				if (it.abilities) {
					absorbs[i] += it.abilities->absorbPercent[i];
				}

        /** Needed because now we also have absorb percentage from imbuing, not only item abilities */
			  for (uint8_t imbuSlot = 0; imbuSlot < it.imbuingSlots; imbuSlot++) {
					uint32_t info = item->getImbuement(imbuSlot);
					if (info >> 8) {
						Imbuement* ib = g_imbuements->getImbuement(info & 0xFF);
						absorbs[i] += ib->absorbPercent[i];
					}
				}
    }
  }
```

Basically the whole for can be replaced by that simple version that I posted.
I'd strongly recommend this to be a method inside items that only retorn the absorb array, it's really bad that it's mixed logic in the protocol part, this should be concerned only about sending messages to client, not doing item attributes and imbuement calculations.

lgrossi: @ArturKnopik can you please do the suggestions above, then we will move with the approval and merge as soon as it's done! Thank you for your collaboration!

src/server/network/protocol/protocolgame.cpp
```diff
@@ -3000,43 +3000,64 @@ void ProtocolGame::sendCyclopediaCharacterCombatStats()
...
+
+		bool skipItemAbility = false;
+
 		if (!it.abilities)
 		{
-			continue;
+			skipItemAbility = true;
 		}
 
 		if (COMBAT_COUNT == 12)
```
lgrossi, line 3019: You don't need this if, you can use the for in this situation also. If combat count is 12 and you add 0 to 11, you can just loop til combat count as the others.

src/server/network/protocol/protocolgame.cpp
```diff
@@ -3000,43 +3000,64 @@ void ProtocolGame::sendCyclopediaCharacterCombatStats()
...
 		{
-			absorbs[0] += it.abilities->absorbPercent[0];
```
ArturKnopik, line 3022: for me this part of code is strange

lgrossi: you don't need this part.. you  can just get rid of it..

This does exactly the same, I don't know why there is an if COMBAT_COUNT == 12, if it does the same as the for loop
```
			for (uint16_t i = 0; i < COMBAT_COUNT; ++i)
			{
				if (it.abilities)
				{
					absorbs[i] += it.abilities->absorbPercent[i];
				}
				absorbs[i] += player->getDeffFromImbuements(indexToCombatType(i), static_cast<slots_t>(slot));
			}
```

xThauma:
You can merge solution for removing all imbuing at once when one in Item expires from my post https://github.com/opentibiabr/otservbr-global/issues/2649 since its imbue issue

majestyotbr:
Need review.

dudantas:
It needs to be reopened in the repository: https://github.com/opentibiabr/canary/

PR #2588, opened 2021-05-16
Convert spells and runes to revscriptsys

beats-dh:
**All player spells and runes have been converted, the monsters spells are still in progress! All need revision.**

**With the contribution of @LeoTK4 @DeroW-FD @Glatharth @APenDel**

### **_Spells:_**

- [x] Attack
- [x] Conjuring
- [x] Healing
- [x] House
- [x] Monster
- [x] Party
- [x] Runes
- [x] Summon
- [x] Support

### **_Fixes:_**
- [x] #923 
- [x] #2506
- [x] #2507 
- [x] #2563

PR #2632, opened 2021-06-12
Prevent players to heal monsters

travisani:


beats-dh:
It has PR #2588 that will change spells to revscript, your modification can be added to it.

ArturKnopik:
use exura gran mas res... still heal monsters, please repair in source instead a single spell

beats-dh:
> use exura gran mas res... still heal monsters, please repair in source instead a single spell

I read you talking about the issue how to do it but I didn't understand very well.

PR #2633, opened 2021-06-12
Update ultimate healing rune.lua

travisani:


beats-dh:
It has PR #2588 that will change spells to revscript, your modification can be added to it.

PR #2635, opened 2021-06-13
Website events.xml update

dex-89:
**How it works?**

1. Move `events.xml` file from `data/XML/events.xml` to website directory
2. Set `eventsUpdate` in `config.lua` as `true`
3. Set corrected path to file from point 1 in `eventsUpdatePath` in `config.lua`
4. Restart your server
5. If all works ok:
image
6. Older `events.xml` in `data/XML` will be overwritten!
7. **If you have multiple servers - all servers will be updated with one file!**
8. Edit only one file (from point 1) to edit events on all servers!
9. If you want disable system - set `eventsUpdate` in `config.lua` as `false`

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -63,6 +63,11 @@ thunderEffect = false
 freeQuests = false
 allConsoleLog = false
 
+-- Update events.xml from website
+-- No support cloudflare redirects and https
+eventsUpdate = false
+eventsUpdatePath = "http://127.0.0.1/events.xml"
+
 -- Deaths
 -- NOTE: Leave deathLosePercent as -1 if you want to use the default
 -- death penalty formula. For the old formula, set it to 10. For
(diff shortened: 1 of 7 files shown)


dudantas:
I don't understand, what is the use of this pull request?

dex-89:
In my spare time I work on the event management system from the website level

dudantas:
> In my spare time I work on the event management system from the website level

That's not what I meant.
It's just that I don't understand exactly what the use of this is. What is the gain in using events on the website instead of the server?

dex-89:
> > In my spare time I work on the event management system from the website level
> 
> That's not what I meant.
> It's just that I don't understand exactly what the use of this is. What is the gain in using events on the website instead of the server?

It's just my idea. It does not need to be implemented

dudantas:
Ok. Thanks for the contribution, but it doesn't interest us because it's bad to create dependency with the site.

PR #2025, opened 2020-11-28
Rework login.lua and added other features

dex-89:
- [X] **Fixed login message**
- - [X] Welcome message added
- [X] **Premium account fixed**
- - [X] Premium Account ends: Teleport to Temple, change addon (citizen) houseless
- [X] **Recruiter system**
How it's working? You need to change the "recruiter" in the database...
The administrator decides what causes the update of this value
- [X] **Moved sendTibiaTime to cpp**
- [X] **Kick other players from account: only one player per account**
if set `onePlayerOnlinePerAccount` as `true`
- [X] **Block kick player if player hasn't PZ, kick the logger**

PR #2642, opened 2021-06-15
Fix connection lost bug

dudantas:


ArturKnopik:
I wonder what it is for, an account with ACCOUNT_TYPE_GOD access is usually an administrator account, the administrator should be able to log in to any number of characters

dudantas:
> I wonder what it is for, an account with ACCOUNT_TYPE_GOD access is usually an administrator account, the administrator should be able to log in to any number of characters

Furthermore, this verification already exists in the source (which exempts the god's account). It was duplicated, so it caused a problem with the connection. It was added in pr https://github.com/opentibiabr/otservbr-global/pull/2025

I merged it, at the time I didn't notice that it could generate the problem, probably because I had the option "onlyOneCharacterPerAccount" as false, so it didn't generate the bug.

majestyotbr (approved):

PR #2672, opened 2021-07-11
Update grand master oberon

vtrolinux:
- Based on the previous script
- Added mechanics identical to the global one
As I didn't find "Holybeam" I kept the same attack but the right one is a straight from san, gg ;)

### DIFF
--- a/data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
+++ b/data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
@@ -1,3 +1,4 @@
+--[[ author vtrolinux(nox) ]]--
 local asking = {
 	[1] = {msg = "You appear like a worm among men!"},
 	[2] = {msg = "The world will suffer for its iddle laziness!"},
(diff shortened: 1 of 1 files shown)


data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
```diff
@@ -1,3 +1,4 @@
+--[[ author vtrolinux(nox) ]]--
 local asking = {
 	[1] = {msg = "You appear like a worm among men!"},
 	[2] = {msg = "The world will suffer for its iddle laziness!"},
```
Deadsmelly: Change "iddle" to "idle"
image

data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
```diff
@@ -1,3 +1,4 @@
+--[[ author vtrolinux(nox) ]]--
```
Glatharth: Would be very kind if you could remove this line.
Since you're not the author of the script and you just made a few alterations on it, it does look like you're claiming credits for the whole thing.
Thanks for your consideration!

data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
```diff
@@ -176,46 +180,58 @@ monster.immunities = {
 	{type = "invisible", condition = true},
 	{type = "bleed", condition = false}
 }
+local controleTempo = 0
```
Glatharth: I think it would be better if we stick to one language only, since everything else is in english, would make more sense.

Schiffers (approved):

Glatharth (changes requested):

data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
vtrolinux: ok

data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
vtrolinux: ok

data/monster/quests/the_secret_library/bosses/grand_master_oberon.lua
vtrolinux: ok

Glatharth (approved):

gabrielscolt0:
Testei aqui e comigo não funcionou, Oberon ficou com vidas infinitas

majestyotbr:
Need review and tests.

dudantas:
It will be fixed in the repository: https://github.com/opentibiabr/otservbr-global-datapack

Issue #2670, opened 2021-07-10
BUG: most of protocolgame.cpp code crash server


gesior:
I just get report about crash from someone on Discord. I think I found reason of it.

OTS multithreading is super simple:
- pass everything to 'dispatcher thread', 'dispatcher' executes one task at once = there is no multithreading problem
- only thing that is not executed in dispatcher thread is 'network packets decrypting and parsing'

#### How to Reproduce?
Run server. Get players. Wait for random crash.

#### Logs or errors
GDB:
```
#0  0x000055ac35bb6e77 in OutputMessage::append(NetworkMessage const&) ()
No symbol table info available.
#1  0x000055ac35bbb027 in ProtocolGame::writeToOutputBuffer(NetworkMessage const&) ()
No symbol table info available.
#2  0x000055ac35bca652 in ProtocolGame::sendResourceBalance(Resource_t, unsigned long) ()
No symbol table info available.
#3  0x000055ac35bca597 in ProtocolGame::sendResourcesBalance(unsigned long, unsigned long, unsigned long) ()
No symbol table info available.
#4  0x000055ac35bcb270 in ProtocolGame::sendMarketEnter(unsigned int) ()
No symbol table info available.
#5  0x000055ac35922233 in Player::sendMarketEnter(unsigned int) ()
No symbol table info available.
#6  0x000055ac35bc470d in ProtocolGame::parseMarketBrowse(NetworkMessage&) ()
No symbol table info available.
#7  0x000055ac35bbc733 in ProtocolGame::parsePacket(NetworkMessage&) ()
No symbol table info available.
#8  0x000055ac35bb6066 in Protocol::onRecvMessage(NetworkMessage&) ()
No symbol table info available.
#9  0x000055ac35ba73d7 in Connection::parsePacket(boost::system::error_code const&) ()
No symbol table info available.
```

Probably 2 threads tried to send packet in same time. They both modified OutputMessage in same time, which result in crash.

#### Expected behavior
'Parse' functions should not execute functions on 'player' (like 'sendMarketEnter'). They should just get data from network packet and pass it to 'dispatcher thread' as 'Task'.

pedriinzk:
`/root/src/src/lua/modules/modules.cpp: In member function ‘void Modules::executeOnRecvbyte(uint32_t, NetworkMessage&, uint8_t) const’:
[...]

PR #2674, opened 2021-07-14
Fix all random crashes generated by protocolgame.cpp #2670

gesior:
This bug: https://github.com/opentibiabr/otservbr-global/issues/2670 should be fixed by rewriting **protocolgame.cpp** and few related files, but I don't have time for it.
If someone have problem with random crashes of otservbr, these changes may fix it.

Bug described above generates absolutely random 'gdb' reports - random function and thread that crashed.

### DIFF
--- a/src/lua/modules/modules.cpp
+++ b/src/lua/modules/modules.cpp
@@ -23,7 +23,9 @@
 
 #include "lua/modules/modules.h"
 #include "creatures/players/player.h"
+#include "game/game.h"
 
+extern Game g_game;
 Modules::Modules() :
 	scriptInterface("Modules Interface")
 {
@@ -95,8 +97,13 @@ Module* Modules::getEventByRecvbyte(uint8_t recvbyte, bool force)
 	return nullptr;
 }
 
-void Modules::executeOnRecvbyte(Player* player, NetworkMessage& msg, uint8_t byte) const
+void Modules::executeOnRecvbyte(uint32_t playerId, NetworkMessage& msg, uint8_t byte) const
 {
+	Player* player = g_game.getPlayerByID(playerId);
+	if (!player) {
+		return;
+	}
+
 	for (auto& it : recvbyteList) {
 		Module module = it.second;
 		if (module.getEventType() == MODULE_TYPE_RECVBYTE && module.getRecvbyte() == byte && player->canRunModule(module.getRecvbyte())) {
--- a/src/lua/modules/modules.cpp
+++ b/src/lua/modules/modules.cpp
@@ -25,7 +25,6 @@
 #include "creatures/players/player.h"
 #include "game/game.h"
 
-extern Game g_game;
 Modules::Modules() :
 	scriptInterface("Modules Interface")
 {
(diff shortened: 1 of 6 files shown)


mattyx14 (approved):

beats-dh:
be careful with the indentation

gesior:
> be careful with the indentation

Removed all formatting changes.

src/lua/modules/modules.cpp
```diff
@@ -95,8 +96,13 @@ Module* Modules::getEventByRecvbyte(uint8_t recvbyte, bool force)
 	return nullptr;
 }
 
-void Modules::executeOnRecvbyte(Player* player, NetworkMessage& msg, uint8_t byte) const
+void Modules::executeOnRecvbyte(uint32_t playerId, NetworkMessage& msg, uint8_t byte) const
 {
+	Player* player = g_game.getPlayerByID(playerId);
```
lgrossi, line 101: Why do you do that? You get the ID from yhe player then retrieve the player again? What is the benefit?

gesior: FIRST: We are in 2 different THREADS. Packet is received in 'main thread' of server and function in modules.cpp is executed in 'dispatcher thread'. We should not pass object references between threads, if we can't guarantee, that object will be still allocated in memory, when second thread start to read/write to it.

Second: It's called with dispatcher = with delay. In meantime player may die and get removed from game. By passing his ID and getting object from game.cpp, I'm checking that he is still in game.

gesior:
Added fixes for few more crash bugs related to **protocolgame.cpp** and multithread programming.

dudantas (approved):

pedriinzk:
/root/src/src/lua/modules/modules.cpp:111:29: error: invalid conversion from ‘Player*’ to ‘uint32_t’ {aka ‘unsigned int’} [-fpermissive]
    module.executeOnRecvbyte(player, msg);
    
 i got it.

dex-89 (approved):
Compilation: no errors (Debian 10)

image

majestyotbr (approved):

Bipbop69 (approved):

majestyotbr:
@lgrossi Pull request waiting for your review.

gesior:
Is it waiting for @lgrossi review or is it expecting me to fix `cpplint` type problem with `NetworkMessage& msg`?
I edited line in `modules.h`, which probably use wrong type, but I did not modify that type.

lgrossi (approved):

src/server/network/protocol/protocolgame.cpp
```diff
@@ -3998,19 +4014,23 @@ void ProtocolGame::sendCoinBalance()
 
 void ProtocolGame::updateCoinBalance()
 {
+	if (!player) {
+		return;
+	}
+
 	g_dispatcher.addTask(
-		createTask(std::bind([](ProtocolGame *client) {
-			if (client && client->player)
```
andersonfaaria, line 4003: @gesior from my side my only question is regarding this line. Since we are not checking for client anymore, can't we have a situation where we could pretend we are another player by changing the reference sent to the function?

gesior: Player can't be 'other player', but player may already use other 'ProtocolGame' (other connection, after player reconnected).
It's possible that we send coin balance to player that did reconnect. After he did reconnect, because code will take 'current connection [protocolgame] of player'.

It's same problem as with whole protocolgame.cpp code, **after someone made 'playerId' based on player GUID, not unique every reconnect** (as in every OTS since 2008). That change was easiest way to make player backpacks and other windows in client 12.x save, but it also created some problems.
In happy scenario, packets send by player before death (like use exura vita), will be handled after he relogin - in case when server is heavy loaded (ex. server save) and player reconnect really quick.
In worst scenario, fast login-logout on same player with few tibia client will crash server - I did not check, how problem of `Player` pointer is handled in game.cpp when playerId is same after every reconnect.

PR #2689, opened 2021-07-28
Gain stamina from training and pz

beats-dh:


data/events/scripts/player.lua
```diff
@@ -507,6 +507,36 @@ end
 function Player:onItemMoved(item, count, fromPosition, toPosition, fromCylinder, toCylinder)
 end
 
+function Player:onChangeZone(zone)
+	if not self:isPremium() then
```
gtkatakura: Could you review the indentation here? It is a bit harder to understand the scopes now.

data/global.lua
```diff
@@ -158,3 +161,32 @@ end
 if not playerDelayPotion then
 	playerDelayPotion = {}
 end
+
+function addStamina(id, amountStamina, delay)
+	local event = staminaRegen[id]
+
+  local player = Player(id)
```
gtkatakura: Here too. Could you review the indentation?

beats-dh:
@gtkatakura Can you check it out now!?

Bipbop69:
a question if the player is with the daily reward that will go up stamina won't they have problems?

Bipbop69 (approved):
is working but it would be nice to get an info message when the player is in pz talking about stamina.

andersonfaaria (approved):

majestyotbr (approved):

majestyotbr:
Tested and working.

data/events/scripts/player.lua
```diff
@@ -507,6 +507,36 @@ end
...
+	  return false
```
gtkatakura: I see that you return booleans here, but seems that it is a void callback, right?
Just to be sure, because if it is the case, you could just do `return` to avoid some confusion here.

beats-dh: This code was taken from otland, I didn't make any changes, I just adapted it to work in otbr and be configurable. I can see this later...

APenDel:
Tested. all good.

mattyx14 (commented):
complete with
https://github.com/opentibiabr/otservbr-global/pull/2692

only training monks jump more to custom ot

mattyx14 (approved):

beats-dh:
> complete with
> #2692
> 
> only training monks jump more to custom ot

Line 164 of global.lua you configure which monster you want to use, it can be any one just change.

mattyx14:
> > complete with
> > #2692
> > only training monks jump more to custom ot
> 
> Line 164 of global.lua you configure which monster you want to use, it can be any one just change.

recommend move to config.lua for more confort

beats-dh (commented):
Ready!

majestyotbr (approved):

PR #2690, opened 2021-07-29
Sort loot of monsters to try most rare itens first

gtkatakura:
# Description

- Sort the loot by chance before register it

### **Actual**

- If you change your `rateLoot` to bigger values, some monsters which have a small container size will most of the time just drop the items in the order they are inserted into the monster.loot, and most of the time this means that rare items don't will drop

### **Expected**

- It should give a chance first for rare items to drop

## Type of change

Please delete options that are not relevant.

  - [x] Bug fix (non-breaking change which fixes an issue)
  - [ ] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] This change requires a documentation update

## How Has This Been Tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

  - [x] Change your `rateLoot` to a big value like 500 and kill an demon, it should drop items with lowest chance of drop (mostly of the times, they are the rare items)

**Test Configuration**:

  - Server Version: 12.64
  - Client: 12.64
  - Operating System: Ubuntu 18.04

## Checklist

  - [x] My code follows the style guidelines of this project
  - [x] I have performed a self-review of my own code
  - [ ] I checked the PR checks reports
  - [x] I have commented my code, particularly in hard-to-understand areas
  - [x] I have made corresponding changes to the documentation
  - [x] My changes generate no new warnings
  - [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/config.lua.dist
+++ b/config.lua.dist
@@ -151,6 +151,13 @@ serverSaveCleanMap = false
 serverSaveClose = false
 serverSaveShutdown = true
 
+-- Sort loot by chance, most rare items drop first
+-- it is good to be setted when you have a higher
+-- rateLoot to avoid losing all rare items when
+-- the corpse size is less than the total of loots
+-- the monster can drop
+sortLootByChance = false
+
 -- Rates
 -- NOTE: rateExp, rateSkill and rateMagic is used as a fallback only
 -- To configure rates see file data/stages.lua
--- a/data/scripts/lib/register_monster_type.lua
+++ b/data/scripts/lib/register_monster_type.lua
@@ -273,8 +273,24 @@ registerMonsterType.events = function(mtype, mask)
 		end
 	end
 end
+
+function sortLootByChance(loot)
+	if not configManager.getBoolean(configKeys.SORT_LOOT_BY_CHANCE) then
+		return
+	end
+
+	table.sort(loot, function(loot1, loot2)
+		if not loot1.chance or not loot2.chance then
+			return 0
+		end
+
+		return loot1.chance < loot2.chance
+	end)
+end
+
 registerMonsterType.loot = function(mtype, mask)
 	if type(mask.loot) == "table" then
+		sortLootByChance(mask.loot)
 		local lootError = false
 		for _, loot in pairs(mask.loot) do
 			local parent = Loot()
@@ -336,6 +352,7 @@ registerMonsterType.loot = function(mtype, mask)
 				parent:setUnique(loot.unique)
 			end
 			if loot.child then
+				sortLootByChance(loot.child)
 				for _, children in pairs(loot.child) do
 					local child = Loot()
 					if children.name then
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -8,6 +8,8 @@ services:
     ports:
       - 7171:7171
       - 7172:7172
+    volumes:
+      - ../data:/otbr/data
     depends_on:
       - mysql
       - website
--- a/src/config/configmanager.cpp
+++ b/src/config/configmanager.cpp
@@ -181,6 +181,9 @@ bool ConfigManager::load()
 	boolean[WEATHER_RAIN] = getGlobalBoolean(L, "weatherRain", false);
 	boolean[WEATHER_THUNDER] = getGlobalBoolean(L, "thunderEffect", false);
 	boolean[ALL_CONSOLE_LOG] = getGlobalBoolean(L, "allConsoleLog", false);
+
+	boolean[SORT_LOOT_BY_CHANCE] = getGlobalBoolean(L, "sortLootByChance", false);
+
 	boolean[FREE_QUESTS] = getGlobalBoolean(L, "freeQuests", false);
 	boolean[SAVE_INTERVAL] = getGlobalBoolean(L, "saveInterval", false);
 	boolean[SAVE_INTERVAL_CLEAN_MAP] = getGlobalBoolean(L, "saveIntervalCleanMap", false);
--- a/src/config/configmanager.h
+++ b/src/config/configmanager.h
@@ -66,6 +66,8 @@ class ConfigManager
 			STAMINA_TRAINER,
 			STAMINA_PZ,
 
+			SORT_LOOT_BY_CHANCE,
+
 			LAST_BOOLEAN_CONFIG /* this must be the last one */
 		};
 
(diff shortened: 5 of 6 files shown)


mrhultin:
I don't see the issue this fixes, these items drop like without this on Global Tibia.

Schiffers:
Why not check the container's size to avoid behavior that you describe? 
Almost all corpses are with the size bigger than 12 slots

andersonfaaria:
I like this idea but I'm afraid of the colateral effect of sorting every loot. This might make the whole server slower because it will be a lot of sorting for each creature dead. Can you add a config in config.lua for it to be turned on/off?

mrhultin:
> I like this idea but I'm afraid of the colateral effect of sorting every loot. This might make the whole server slower because it will be a lot of sorting for each creature dead. Can you add a config in config.lua for it to be turned on/off? 

This only happens when monsters are registered, do should only happen on launch and/or reloads.

andersonfaaria:
Can you show me a comparison of startup time with and without this PR then?

mrhultin:
> Can you show me a comparison of startup time with and without this PR then? 

Just check it out and run it yourself, I cannot imagine we're talking more than a second or two even on a slow cpu.

andersonfaaria:
> > Can you show me a comparison of startup time with and without this PR then?
> 
> Just check it out and run it yourself, I cannot imagine we're talking more than a second or two even on a slow cpu.

Unfortunately I don't have a server to test, in the ideal world I'd like a config option for this, it's hard to measure the impact besides startup:

a) How this affects auto-loot and reward chest?
b) What about servers that can't prohibit bot, wouldn't that be easier for the bots to pick better equipments?

If OP can rebase considering a turn on/off switch in config.lua I can merge it confidently, other than that I'd prefer to wait a few other opinions/data.

APenDel:
PR coming soon in the day, changed all the rate-skill/exp/loot/spawn, for events.xml.

based and compared with official tibia.

gesior:
> a) How this affects auto-loot and reward chest?

Should not affect at all. Auto-loot messages may change order of items, as we change it in monster.
> b) What about servers that can't prohibit bot, wouldn't that be easier for the bots to pick better equipments?

It may be harder to loot for players **without bot**, but only on servers with XXX loot rate, where monsters drop a lot of items. Rare items will be at end of corpse container - we add them first to corpse, so they will be at end of it after adding more items.

> Can you show me a comparison of startup time with and without this PR then?

We are talking about sorting 1491 (monsters count in `data/monsters`) tables, each with 5-20 items. It should take 0.01 sec.

gesior (approved):

APenDel:
mmm, i think if this is for a custom server when use rate XXXX, this need a config value too...
if i don't want use the sort method?. (I haven't seen the files, I don't know if it has config or not).

gtkatakura:
Do monsters have a known order of loot to be applied? When people add some loot to some monster, the order is based on some information from tibia global or it is just randomly inserted into the table of loot?

gtkatakura:
>>> b) What about servers that can't prohibit bot, wouldn't that be easier for the bots to pick better equipment?

About dropping more rare loots, it only affects when you had a small container size and you already fulfilled the container. When you have a lower rate loot, it doesn't make difference because it is harder to fulfill the container with normal items before trying to apply the chance over rare items. Also, monsters like dragon lord and others that have a big container size only impact the order of the items showed into the container because they have a container size that gives a chance to try every loot item.

Of course, would be nice if we have some kind of automated tests to detect the changes over the loot to confirm this, instead of manual tests like I did ;/

In the future we can try to make some kind of automated tests for loot ;D

gtkatakura:
>>> It may be harder to loot for players without bot, but only on servers with XXX loot rate, where monsters drop a lot of items. Rare items will be at end of corpse container - we add them first to corpse, so they will be at end of it after adding more items.

To me, this seems the most relevant impact.

I see you guys are very concern and because we don't have automated tests to check the real impacts of this change over all monsters loot, I can create a variable to enable/disable it (and be disable by default), but seems that maybe I can just check the rateLoot (like only enable for rateLoot>30/50)?

beats-dh:
> > > > It may be harder to loot for players without bot, but only on servers with XXX loot rate, where monsters drop a lot of items. Rare items will be at end of corpse container - we add them first to corpse, so they will be at end of it after adding more items.
> 
> To me, this seems the most relevant impact.
> 
> I see you guys are very concern and because we don't have automated tests to check the real impacts of this change over all monsters loot, I can create a variable to enable/disable it (and be disable by default), but seems that maybe I can just check the rateLoot (like only enable for rateLoot>30/50)?

Leave configurable by config.lua, this should be the correct behavior for anything outside of Tibia Rl

andersonfaaria:
> Should not affect at all. Auto-loot messages may change order of items, as we change it in monster.

From what I can recall, reward chest picks the loot from corpse and re-distribute to players following a logic to distribute the loot among the participants.
I can't tell right now the logic, but having the loot sorted might increase the distribution of best value loot in an unplanned way.

Of course I'm just thinking this deeply, we should have ways to foresee this by reviewing how those mechanics work.

if for each loot I have a % of attributing it to someone, you agree that the last user percentage of getting the item is not just his percentage but the combination of all other users that had 1-% multiplied?

I'm giving you very specific and rare situations on where having the loot sorted could make such systems behave differently, but we need to make this decision concious about the impacts, going beyond what we think "should" or "shouldn't" happen.

Don't get me wrong, probably you're right and it shouldn't impact, but having the configuration to turn it on/off make it safer in case we spot a problem when releasing it mass population that will be testing this in "prod" environments.

gesior:
> From what I can recall, reward chest picks the loot from corpse and re-distribute to players following a logic to distribute the loot among the participants.
> I can't tell right now the logic, but having the loot sorted might increase the distribution of best value loot in an unplanned way.

Just checked logic. Reward chest does not generate boss loot and distribute to players. It generates new list of loot for every player based on 'loot chance' (monster loot config) multiplied by given player 'loot factor'. If there is one 'magic plate armor' on boss loot list, all players may get it. Boss will drop 10 MPA, one for each player - if they are all lucky.
Order of items, number of items and order of players [from high to low 'score' in fight] do not affect reward chest system.

andersonfaaria (approved):
based on Gesior's analysis, it's approved from my side.
I'd still be in favor of making it configurable on config.lua but we can do it in another PR.

majestyotbr:
Tested and working. Just need to leave configurable by config.lua.

gtkatakura:
@andersonfaaria @MajestyOTBR done

majestyotbr (approved):

andersonfaaria (approved):
approved, thanks for making it configurable.
I'd even put it as default to true but that's fine

PR #2693, opened 2021-08-04
Add setPzLocked spell function

dudantas:
# Description
Work with: https://github.com/opentibiabr/otservbr-global/pull/2588

### DIFF
--- a/src/lua/scripts/luascript.cpp
+++ b/src/lua/scripts/luascript.cpp
@@ -17999,6 +17999,19 @@ int LuaScriptInterface::luaSpellVocation(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaSpellSetPzLocked(lua_State* L)
+{
+	// spell:setPzLocked(bool)
+	Spell* spell = getUserdata<Spell>(L, 1);
+	if (spell) {
+		pushBoolean(L, spell->setPzLocked(getBoolean(L, 2));
+		}
+	} else {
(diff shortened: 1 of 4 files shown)


src/lua/scripts/luascript.cpp
```diff
@@ -17999,6 +17999,19 @@ int LuaScriptInterface::luaSpellVocation(lua_State* L)
 	return 1;
 }
 
+int LuaScriptInterface::luaSpellSetPzLocked(lua_State* L)
+{
+	// spell:setPzLocked(bool)
+	Spell* spell = getUserdata<Spell>(L, 1);
+	if (spell) {
+		pushBoolean(L, spell->setPzLocked(getBoolean(L, 2));
```
beats-dh: One is missing ) after the 2. But when trying to compile we get the error " Cannot convert an argument 2 from 'void' to 'bool' "

src/lua/scripts/luascript.cpp
```diff
@@ -17999,6 +17999,19 @@ int LuaScriptInterface::luaSpellVocation(lua_State* L)
...
+		}
```
beats-dh: To remove }

beats-dh (changes requested):

Beats-Dh:
in runes is ```rune:setPzLocked()``` takes that add tbm in src?

dudantas:
https://github.com/Beats-Dh/otservbr-global/pull/51

PR #2692, opened 2021-08-01
Some fixes events, configkey, optional stages, xp gain rate

APenDel:
### **Actual**
Issues opentibiabr/otservbr-global#2464

On login you have one "XP Gain Rate", when you get experience this value changed.
Event schedule only display info, but doesn't work.

### **Expected**
[Formulae Tibia Wiki](https://tibia.fandom.com/wiki/Formulae)
  - [x] Display correctely "XP Gain Rate".
  - [x] Event Schedule server-side.
  - [x] Exercise Training get bonus of events.

### global.lua & statup.lua
- moved form `global.lua -> globalevents\others\startup.lua` all events rates variables.
`getEventSSkill, getEventSExp and getEventSLoot` now load after events module.
- set `GlobalStorage.XpDisplayMode -> 1`

### player.lua
- Fixed `onGainExperience`; Added `preyBonus & staminaBoost` variable.
now use simple rule of 3...
Rat: if 5 exp --- 100% | ? exp --- 150% so... 5 / 100 * 150 = 7(BaseXpGain + etc) see [Rat](https://www.tibiawiki.com.br/wiki/Rat)

### lib\core\functions.lua
- `getLootRandom` approximate to real tibia.
- `getRateFromTable` when doesn't have stages, return default.

### exercise_training.lua
- deleted unnecessary variables; if not reused.
- added `SCHEDULE_SKILL_RATE` to calculate exp.

### configmanager.cpp & configmanager.h
- added `RATE_USE_STAGES` boolean config `rateUseStages = true`.
- fixed `RATE_EXPERIENCE`.
- set base rates to 1.

## creaturescript\others\login.lua
- fixed `setBaseXpGain`, now work with events too.

## spawn.cpp & spawn.h
- fixed interval spawn time.
- when the spawn time is less than `MINSPAWN_INTERVAL`, addMonster with min value.
- now `interval` to `30000`, less check time.

### game.cpp, scripts\luascript.cpp & protocolstatus.cpp
- some fixes

### src\otserv.cpp & config.lua.dist
- On/Off stagesExp(def: true).
- set rates to 1

### events.xml
- changed date for test.
- can use exprate="0", with any argument (this reduce the rate to 0, but not the exp).
NOTE: exprate, lootrate, spawnrate or skillrate in 100, it's default value.

## Type of change
  - [x] Bug fix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)
[...]

PR #2706, opened 2021-08-18
Some fixes events, configkey, optional stages, xp gain rate

beats-dh:
PR #2692

### **_Fixes:_**
- [x] #2464


majestyotbr (approved):

mattyx14 (approved):

Bipbop69 (commented):
player.lua:Player@onGainExperience
data/events/scripts/player.lua:744: attempt to perform arithmetic on global 'baseExp' (a nil value)
stack traceback:
        [C]: in function '__mul'
        data/events/scripts/player.lua:744: in function <data/events/scripts/player.lua:702>
not doing the calculation when player has exp play + xp store + green stamina

beats-dh:
> player.lua:Player@onGainExperience
> data/events/scripts/player.lua:744: attempt to perform arithmetic on global 'baseExp' (a nil value)
> stack traceback:
> [C]: in function '__mul'
> data/events/scripts/player.lua:744: in function <data/events/scripts/player.lua:702>
> not doing the calculation when player has exp play + xp store + green stamina

It's accusing error on line 744 I checked and didn't make changes to it, you have the PR Branch without changes?

Bipbop69:
> It's accusing error on line 744 I checked and didn't make changes to it, you have the PR Branch without changes?

I'm using the current version with this commit.
putting it uses the exp of the stage in place of the base is ok.

but you will have the problems I said

exp prey is not being and store is not pulling correctly
in this case it would have to be applied in the stage the player is in but it is always being applied in the mob's exp

dudantas (approved):

dudantas:
It will be fixed in the repository: https://github.com/opentibiabr/otservbr-global-datapack

PR #2710, opened 2021-08-27
Fix decaying items crash in bosses #2709

### LINKED ISSUE #2709: [Bug]: Bosses drop decaying items to Reward Box (crash)
### What happened?

Bosses drop decaying items (version where they are decaying, like equipped rings) to reward container.
When you open boss corpse and reward container, that reward container has `parent` set to `corpse item`, but `corpse` is not informed about new item inside:
https://github.com/opentibiabr/otservbr-global/blob/develop/src/lua/creature/actions.cpp#L431

When corpse disappear (decay to 0 over time), it does not remove reward container - which is good, as we don't want to remove items.. that are the same items as you see when you open 'reward chest' later.
Problem is that `decaying ring` still has pointer to `parent`, which is removed from RAM and that ring is still on list of items `toDecay`. Within second executes `Game::checkDecay` and calls `canDecay` on Item*, which goes to parent->parent->parent... of item and checks, if it's not null.... and IT'S NOT NULL, **it's pointing to removed corpse, so it crashes server.**

It's very rare. On popular server gives few crashes a week, but it still should be fixed.

Example of boss with invalid item id: item ID 26184 in Maxxenius
https://github.com/opentibiabr/otservbr-global/blob/develop/data/monster/quests/the_dream_courts/bosses/maxxenius.lua#L89
You can check in items.xml that it's version of 'collar of red plasma' that you get AFTER equipping item.

### How to Reproduce?

1. Kill some boss that drop item with invalid ID. Invalid ID is ID that points to item in state 'decaying', not version 'before equip'. Sometimes it will spawn in boss in state in which it will count down items time left (duration), as it decays.
2. Open corpse, open 'reward container' and make sure it's decaying. It's required, as bug happens, when you open 'reward container' and see list of loot.
3. Stay online on player who opened 'reward container' until corpse disappear.
4. Done. It will crash - probably sometimes, not every time.

In `gdb` report you will see:
```
#1  0x000055bff0f23f7e in Item::isRemoved (this=0x7fdf3d5c1b60)
    at /home/abcd/src/items/item.h:1071
No locals.
#2  0x000055bff0f23f7e in Item::isRemoved (this=0x7fe02cae8440)
    at /home/abcd/src/items/item.h:1071
```
as it checks `item->isRemoved()`, which calls `item->parent->isRemoved()`. So there will be 2 calls to `Item::isRemoved`.

### Relevant log output

```shell
#1  0x000055bff0f23f7e in Item::isRemoved (this=0x7fdf3d5c1b60)
    at /home/abcd/src/items/item.h:1071
No locals.
#2  0x000055bff0f23f7e in Item::isRemoved (this=0x7fe02cae8440)
    at /home/abcd/src/items/item.h:1071
No locals.
#3  0x000055bff0f79a13 in Item::canDecay (this=0x7fe02cae8440)
    at /home/abcd/src/items/item.cpp:2277
        it = @0x7fe084c73a98: {group = 1223914544, type = 32736, id = 15024,
          clientId = 33991, stackable = 224, isAnimation = 127,
          isPodium = false,

(...)

#4  0x000055bff0ed9b78 in Game::checkDecay (this=0x55bff1a3a200 <g_game>) at /home/abcd/src/game/game.cpp:6529
        item = 0x7fe02cae8440
        duration = 529000
        decreaseTime = 1000
        __func__ = "checkDecay"
        bucket = 1
        it = 0x7fe02cae8440
        end = 0xaea
#5  0x000055bff0f0fd67 in std::__invoke_impl<void, void (Game::*&)(), Game*&> (__f=<optimized out>, __t=<optimized out>)
    at /usr/include/c++/8/bits/invoke.h:73
```

### Attach any relevant file

_No response_

### Version

12.x (Default)

### What OS are you seeing the problem on?

Linux

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

gesior:
# Description

Bosses drop decaying items. If you open 'reward container' in boss corpse, not pick decaying items and corpse disappear, it will crash server.

### **Actual**

Server crash.

### **Expected**

Server does not crash.

## Fixes

#2709

## Type of change

  - Bug fix (non-breaking change which fixes an issue)
Operating System:

### DIFF
--- a/data/lib/others/reward_boss_lib.lua
+++ b/data/lib/others/reward_boss_lib.lua
@@ -66,7 +66,7 @@ function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
 	if decayTo and decayTo >= 0 and decayTime and decayTime ~= 0 then
 		local transformDeEquipId = itemType:getTransformDeEquipId()
 		if transformDeEquipId and transformDeEquipId > 0 then
-			print("[Warning - MonsterType.createLootItem] Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
+			Spdlog.warn("[MonsterType.createLootItem] - Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
 			lootBlock.itemId = transformDeEquipId
 		else
 			print("[Error - MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
--- a/data/lib/others/reward_boss_lib.lua
+++ b/data/lib/others/reward_boss_lib.lua
@@ -69,7 +69,7 @@ function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
 			Spdlog.warn("[MonsterType.createLootItem] - Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
 			lootBlock.itemId = transformDeEquipId
 		else
-			print("[Error - MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
+			Spdlog.error("[MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
 			return lootTable
 		end
 	end
--- a/data/lib/others/reward_boss_lib.lua
+++ b/data/lib/others/reward_boss_lib.lua
@@ -60,6 +60,20 @@ function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
 		end
 	end
 
+	local itemType = ItemType(lootBlock.itemId)
+	local decayTo = itemType:getDecayId()
+	local decayTime = itemType:getDecayTime()
+	if decayTo and decayTo >= 0 and decayTime and decayTime ~= 0 then
+		local transformDeEquipId = itemType:getTransformDeEquipId()
+		if transformDeEquipId and transformDeEquipId > 0 then
+			print("[Warning - MonsterType.createLootItem] Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
+			lootBlock.itemId = transformDeEquipId
+		else
+			print("[Error - MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
+			return lootTable
+		end
+	end
+
 	while itemCount > 0 do
 		local n = math.min(itemCount, 100)
 		itemCount = itemCount - n


mrhultin:
I see the return, and unable to test, but wouldnt this make the reward container empty if the first item has decay or do I need more coffee here?

gesior:
@mrhultin 
Probably more coffee :)
`MonsterType.createLootItem(self, lootBlock, chance, lootTable)` is called for every loot item in boss:
https://github.com/opentibiabr/otservbr-global/blob/cb3e8408dbd2fa6c78a7d5ea3ba83649086ce80c/data/lib/others/reward_boss_lib.lua#L96

It should block only single bugged items.

mrhultin:
> @mrhultin 
> 
> Probably more coffee :)
> 
> `MonsterType.createLootItem(self, lootBlock, chance, lootTable)` is called for every loot item in boss:
> 
> https://github.com/opentibiabr/otservbr-global/blob/cb3e8408dbd2fa6c78a7d5ea3ba83649086ce80c/data/lib/others/reward_boss_lib.lua#L96
> 
> 
> 
> It should block only single bugged items.

Yeah you are correct, ofc. I like this fix, its elegant!

ArturKnopik:
I prefer check such things  during startup(check during adding loot to monste[boss] and change/remove decaying items), we don't perform any additional operation during normal gameplay #optymization_lover

mattyx14 (approved):

Bipbop69 (approved):

data/lib/others/reward_boss_lib.lua
```diff
@@ -60,6 +60,20 @@ function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
 		end
 	end
 
+	local itemType = ItemType(lootBlock.itemId)
+	local decayTo = itemType:getDecayId()
+	local decayTime = itemType:getDecayTime()
+	if decayTo and decayTo >= 0 and decayTime and decayTime ~= 0 then
+		local transformDeEquipId = itemType:getTransformDeEquipId()
+		if transformDeEquipId and transformDeEquipId > 0 then
+			print("[Warning - MonsterType.createLootItem] Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
```
dudantas: You can change here to 
```suggestion
			Spdlog.warn("[MonsterType.createLootItem] - Convert boss '" .. self:name() .. "' reward ID '" .. lootBlock.itemId .. "' to ID " .. transformDeEquipId .. ".")
```

data/lib/others/reward_boss_lib.lua
```diff
@@ -60,6 +60,20 @@ function MonsterType.createLootItem(self, lootBlock, chance, lootTable)
...
+			lootBlock.itemId = transformDeEquipId
+		else
+			print("[Error - MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
```
dudantas:
```suggestion
			Spdlog.error("[MonsterType.createLootItem] Cannot add item " .. lootBlock.itemId .. " as boss " .. self:name() .. " reward. It has decay.")
```

dudantas (changes requested):
Your pull is very important. Thanks for the contribution. I left two suggestions, as our source was changed logs to SPDLOG.

dudantas (approved):

majestyotbr (approved):

PR #2173, opened 2020-12-27
Fix summons & familiars

dex-89:
- [X] Changed storage name
- [X] Fixed event 'onLook'
- [X] Added missing line (#2158) : by @hyresu 
- [X] Fixed familiars disappearing
- [X] Player can attack his familiar / summon
- [X] Fixed event 'onDeath'
- [X] Disappearance communication
- [X] Familiars fly in front of player: by @hyresu 
- [X] Summons with targetdistance running away from target (https://github.com/opentibiabr/otservbr-global/issues/1250): by @hyresu
- [X] Fixed Dawnport crash @hyresu 
- [X] Fixed Avoid unnecessary heap allocations in function "ProtocolGame::sendOutfitWindow()" @SaiyansKing 
- [X] Fixed: `he gets trapped by monsters the summon disappear`

PR #2717, opened 2021-09-05
Fix summons and familiars bug

dudantas:
# Description

The familiars kept following the glued players, the summon icon didn't appear properly. These problems appear after the pull https://github.com/opentibiabr/otservbr-global/pull/2173

I reverted most of the changes related to summons and family members, because the code was wrong, the master is passed in Game.createMonster() and not in creature:addSummon(), this second is only used with summons from monsters and in the convince creature.

### **Actual**
image

### **Expected**
image

## Type of change

Please delete options that are not relevant.

  - [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

  - [x] Create a summon with the icon (shield)
  - [x] Create a summon and this is following with a distance of 1 sqm.

**Test Configuration**:

  - Server Version: 12.64
  - Client: 12.64
  - Operating System: Windows

## Checklist

  - [x] My code follows the style guidelines of this project
  - [x] I have performed a self-review of my own code
  - [x] I checked the PR checks reports
  - [x] I have commented my code, particularly in hard-to-understand areas
  - [x] I have made corresponding changes to the documentation
  - [x] My changes generate no new warnings
  - [x] I have added tests that prove my fix is effective or that my feature works


data/scripts/spells/summon/druid_familiar.lua
```diff
@@ -47,26 +47,25 @@ function spell.onCastSpell(player, variant)
...
-		player:sendCancelMessage(RETURNVALUE_NOTPOSSIBLE)
-		player:getPosition():sendMagicEffect(CONST_ME_POFF)
+		player:sendCancelMessage(RETURNVALUE_NOTENOUGHROOM)
+		playerPosition:sendMagicEffect(CONST_ME_POFF)
 		return false
 	end
 
-	player:addSummon(myFamiliar)
 	myFamiliar:setOutfit({lookType = player:getFamiliarLooktype()})
-	--myFamiliar:reload()
```
dex-89, line 65: uncomment this line in the spell file and a similar in creaturescripts/familiars's file and it will appear icons

Check all vocations, do summons work in accordance with global Tibia...

I was working on this problem, I was struggling with the fact that 2 familiars act differently and the other two act differently (I can't remember which ones are which)

dudantas: > uncomment this line in the spell file and a similar in creaturescripts/familiars's file and it will appear icons
> 
> Check all vocations, do summons work in accordance with global Tibia...
> 
> I was working on this problem, I was struggling with the fact that 2 familiars act differently and the other two act differently (I can't remember which ones are which)

The behavior is not correct. The summons are following the master (as if they had auto follow). Inside other small problems. I believe there is a better solution. Furthermore, the correct thing is to pass the master (player) in Game.createMonster(), that addSummon function is not being used correctly. If you look at the saiyans base, where this was initially implemented, it doesn't use it on summons (utevo res), on summons it uses Game.createMonster(), it uses "player:addSummon" only on monster summons and in the convince creature (precisely because you don't need the shield).

Players summons:
https://github.com/dudantas/optimized_forgottenserver/blob/0ed972e420dfc22afaf901568a5821c4602b8429/data/spells/scripts/support/summon_creature.lua#L38

Monsters summons: https://github.com/dudantas/optimized_forgottenserver/blob/0ed972e420dfc22afaf901568a5821c4602b8429/data/scripts/lib/register_monster_type.lua#L148

SaiyansKing: > If you look at the saiyans base, where this was initially implemented, it doesn't use it on summons (utevo res), on summons it uses Game.createMonster(), it uses "player:addSummon" only on monster summons and in the convince creature (precisely because you don't need the shield).

I make createMonster that way because I've made some heavy changes to setMaster function so it'll properly make client recognize summons but since it is heavy and not required in most casses I simply modified the createMonster function to apply master before the monster is placed on map to not waste CPU.

https://github.com/dudantas/optimized_forgottenserver/blob/0ed972e420dfc22afaf901568a5821c4602b8429/src/luascript.cpp#L7730
It now refresh monster target/friend list(I was surprised that it wasn't like that by default) and make use of new function "updateCreatureData" which works similar to the "reloadCreature" otservbr have but it refresh cached data for all players that knows the creature(it is required for summon "icons" to work) not only for some spectators.

dudantas: > > If you look at the saiyans base, where this was initially implemented, it doesn't use it on summons (utevo res), on summons it uses Game.createMonster(), it uses "player:addSummon" only on monster summons and in the convince creature (precisely because you don't need the shield).
> 
> I make createMonster that way because I've made some heavy changes to setMaster function so it'll properly make client recognize summons but since it is heavy and not required in most casses I simply modified the createMonster function to apply master before the monster is placed on map to not waste CPU.
> 
> https://github.com/dudantas/optimized_forgottenserver/blob/0ed972e420dfc22afaf901568a5821c4602b8429/src/luascript.cpp#L7730
> It now refresh monster target/friend list(I was surprised that it wasn't like that by default) and make use of new function "updateCreatureData" which works similar to the "reloadCreature" otservbr have but it refresh cached data for all players that knows the creature(it is required for summon "icons" to work) not only for some spectators.

Yes, the way you did it is great. The problem is that in my view, the pull request in question (which I mentioned in my pull comment) mixed two behaviors, which was the way you did it and the way it was already here.

dex-89: True, they follow the master so artificially. In my free time, I will check what it looks like after your changes and let you know

dudantas: Beauty. If you want to upload another request, feel free. Regards!

dudantas: Ok! If you want to send another pull request, feel free. Regards!

dex-89: It will be not better? (line with blue balls) file spells/summons
image

majestyotbr (approved):

PR #2726, opened 2021-09-06
Freequest freeze

Glatharth:
The cause of the freeze was the massive amount of setPlayerStorage() being executed, I transferred this task to the database, when a char is created it will execute a procedure that will add all storages.

The choice of procedure is to walk and start a pattern where the procedure creation is more normal, so that we can optimize systems that require a large amount of executions in the database, it will also serve for sites to communicate in a more friendly.


dex-89:
This is not a good solution because freequests are related to config.lua (freequests = true/false). Not everyone wants to give access to all quests on their server

Glatharth:
Yes, I thought about that too, I'll make a table where it will contain all the storages the server owner wants, if he doesn't want any, he can leave the trigger and the procedure aside or simply don't insert any storage in the table.
I don't think it's a bad solution just because there's "no way" to leave false on freequests. I think it should just be adapted

dudantas:
Let's think of a better system, shall we? Then we can open a pull request in the new repository: https://github.com/opentibiabr/otservbr-global-datapack/

I'm closing, because migrations are only executed once, then this would generate a problem: when a new storage is added there, how will it run? And considering that a new migration will be added (it will eventually be), it won't be possible to run the migrations again, it will be necessary to create another one.

But, thanks!

PR #2734, opened 2021-09-09
Update 10 12x with develop

DamianDEB:
Hi how i can disabled client 12 and other thing's i wanna work only on Tibia 10 
second question this source have big problem with stability and exploit ?


beats-dh:
ask on https://discordapp.com/invite/3NxYnyV

PR #2744, opened 2021-09-15
Convert datapack to use Client IDs

dmarszk:
# Description

Migrate away from Server IDs. The reasoning behind not using Server IDs for items anymore is making it much more maintainable and in-line with how global servers do it.

## TODO
  - [ ] Validate that no functionality is broken
  - [ ] Ensure no storage IDs got modified
  - [ ] Ensure that no `items.xml` changes from global are lost
  - [ ] Rebase on top of canary npcs adaptation

## Conversion script used (or some draft of it)
https://gist.github.com/dmarszk/2fe6ec3b6f4a2f868500858b2174068f


dmarszk:
I have exploded it into a few separate commits to make the review slightly easier.

dudantas:
Added to: https://github.com/opentibiabr/otservbr-global-datapack/pull/1
Thanks!

